PR #2, opened 2015-04-04
[OSJC-150] Move to new package structure of com.openshift.restclient, re...

jcantrill:
...move v2 classes


brenton:
@danmcp, we probably want to keep a branch around for OSE 2.x and Online that still has the v2 code, right?

danmcp:
@brenton v2 is still under: https://github.com/openshift/openshift-java-client/

brenton:
@danmcp, ahh, that's right.  Would you mind if I crosslink the repos in the README to hopefully save someone some time in the future?

PR #7, opened 2015-04-15
[JBIDE-19601] Status kind being returned with incorrect version and clie...

jcantrill:
...nt unable to recognize it


jcantrill:
@adietish check it out

adietish:
this is a bug in the server, it's responding in v1beta3 even though we're requesting v1beta1: https://github.com/GoogleCloudPlatform/kubernetes/issues/6876.
closing without merging.

PR #10, opened 2015-04-21
[OSJC-157] Throw unauthorized exception when a 403 status is returned

jcantrill:
@adietish found our issue when reconnecting after eclipse session restart.  Per my discussion with @maxandersen , im putting this into master


src/main/java/com/openshift/internal/restclient/KubernetesAPIVersion.java
```diff
@@ -15,7 +15,8 @@
  * @author Jeff Cantrill
  */
 public enum KubernetesAPIVersion implements APIModelVersion{
-	v1beta1(1);
+	v1beta1(1),
```
adietish: I am pretty convinced that using an enum for api version is not the right tool at hand. Enums typically are the right tool for closed exhaustive choice whereas in API versions we clearly have an open ended choice. I would clearly switch to a class and move away from enum.

PR #11, opened 2015-04-22
[OSJC-154] Add capability to process and apply templates for a given pro...

jcantrill:
...ject

[OSJC-158] Allow CapabilityVisitor to return a value
[OSJC-159] Projects return their name when namespace is empty
[OSJC-160] Allow getting status from Status Kind

@adietish Please review.  Supports being able to create app from template

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/DefaultClient.java
+++ b/src/main/java/com/openshift/internal/restclient/DefaultClient.java
@@ -8,7 +8,7 @@
  ******************************************************************************/
 package com.openshift.internal.restclient;
 
-import static com.openshift.restclient.capability.CapabilityInitializer.initializeClientCapabilities;
+import static com.openshift.internal.restclient.capability.CapabilityInitializer.initializeClientCapabilities;
 
 import java.net.MalformedURLException;
 import java.net.SocketTimeoutException;
@@ -249,12 +249,13 @@ public class DefaultClient implements IClient, IHttpStatusCodes{
 	
 	@SuppressWarnings("unchecked")
 	@Override
-	public <T extends ICapability> void accept(CapabilityVisitor<T> visitor){
+	public <T extends ICapability, R> R accept(CapabilityVisitor<T, R> visitor, R unsupportedCapabililityValue){
 		if(!capabilitiesInitialized) initializeCapabilities();
 		if(capabilities.containsKey(visitor.getCapabilityType())){
 			T capability = (T) capabilities.get(visitor.getCapabilityType());
-			visitor.visit(capability);
+			return (R) visitor.visit(capability);
 		}
+		return unsupportedCapabililityValue;
 	}
 
 	public List<KubernetesAPIVersion> getKubernetesVersions() {
--- a/src/main/java/com/openshift/internal/restclient/capability/resources/ProjectTemplateListCapability.java
+++ b/src/main/java/com/openshift/internal/restclient/capability/resources/ProjectTemplateListCapability.java
@@ -1,0 +1,44 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Red Hat, Inc. Distributed under license by Red Hat, Inc.
+ * All rights reserved. This program is made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution, and is
+ * available at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors: Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package com.openshift.internal.restclient.capability.resources;
+
+import java.util.Collection;
+
+import com.openshift.restclient.IClient;
+import com.openshift.restclient.ResourceKind;
+import com.openshift.restclient.capability.resources.IProjectTemplateList;
+import com.openshift.restclient.model.IProject;
+import com.openshift.restclient.model.template.ITemplate;
+
+public class ProjectTemplateListCapability implements IProjectTemplateList {
+	
+	private IProject project;
+	private IClient client;
+
+	public ProjectTemplateListCapability(IProject project, IClient client) {
+		this.project = project;
+		this.client = client;
+	}
+
+	@Override
+	public boolean isSupported() {
+		return client != null && project != null;
+	}
+
+	@Override
+	public String getName() {
+		return this.getClass().getSimpleName();
+	}
+
+	@Override
+	public Collection<ITemplate> getTemplates() {
+		return client.<ITemplate>list(ResourceKind.Template, project.getNamespace());
+	}
+
+}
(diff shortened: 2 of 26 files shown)


src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -249,12 +249,13 @@ public  boolean supports(Class<? extends ICapability> capability) {
 	
 	@SuppressWarnings("unchecked")
 	@Override
-	public <T extends ICapability> void accept(CapabilityVisitor<T> visitor){
+	public <T extends ICapability, R> R accept(CapabilityVisitor<T, R> visitor, R unsupportedCapabililityValue){
```
adietish: just so that I understand the API: what is the aim of unsupportedCapabilityValue?

src/main/java/com/openshift/internal/restclient/capability/resources/ProjectTemplateListCapability.java
```diff
@@ -0,0 +1,62 @@
...
+	private IProject project;
+	private IClient client;
+
+	public ProjectTemplateListCapability(IProject project, IClient client) {
+		this.project = project;
+		this.client = client;
+	}
+
+	@Override
+	public boolean isSupported() {
```
adietish: do I get this right? the check whether the protocol version we're talking (v1betaX) is implicit here. The capability wouldnt be present if it wasnt. So the only further check to know if this is supported is the presence of a client (is that possible?) and a non-null project. Right?

src/main/java/com/openshift/internal/restclient/capability/resources/ProjectTemplateListCapability.java
```diff
@@ -0,0 +1,62 @@
...
+		return this.getClass().getSimpleName();
+	}
+
+	@Override
+	public Collection<ITemplate> getTemplates() {
+		Collection<ITemplate> all = new ArrayList<ITemplate>();
+		try {
+			all.addAll(client.<ITemplate>list(ResourceKind.Template, DEFAULT_NAMESPACE));
+		}catch(UnauthorizedException e) {
+			LOG.warn("Unauthorized to list templates in the default namespace", DEFAULT_NAMESPACE, e);
```
adietish: Shouldnt exceptions bubble up to the caller and "interrupt" his call?

src/main/java/com/openshift/internal/restclient/model/template/Template.java
```diff
@@ -56,4 +56,14 @@ public Template(ModelNode node, IClient client, Map<String, String []> propertyK
 		return resources;
 	}
 
+	@Override
+	public void updateParameterValues(Collection<IParameter> parameters) {
+		Map<String, IParameter> actuals = getParameters();
+		for (IParameter param : parameters) {
+			if(actuals.containsKey(param.getName())) {
```
adietish: shouldnt we somehow tell the user that some parameters were not found and thus not subsituted?

src/main/java/com/openshift/internal/restclient/DefaultClient.java
jcantrill: @adietish needed the ability to return a value from the visitor, but also handle the case where the visitor wasn't 'visited'.  Without providing a 'default' value you can run into either null pointers our class cast exceptions.  It's not as pretty as it used to be :(

adietish: ok got it, it's a default value.
still pretty slick imho.

src/main/java/com/openshift/internal/restclient/capability/resources/ProjectTemplateListCapability.java
jcantrill: @adietish I'm not sure I fully understand what you are asking.  It is assumed that we can talk to whatever version of the server we are connected.  Conceptually, you could ask the client the api model version as I think we expose it.  The additional non-null checks are really just sanity checks as they should never be null.  The only case I can think off is if this class is instantiated outside of using our connection logic.

src/main/java/com/openshift/internal/restclient/capability/resources/ProjectTemplateListCapability.java
jcantrill: @adietish This I wasn't really certain how to handle unless we did so with multiple calls from jbosstools.  You should be able to get templates from your project and the openshift namespace, but I didn't want to kill the flow if the policy wasnt set-up for you to retrieve from the openshift namespace.  suggestions?

src/main/java/com/openshift/internal/restclient/model/template/Template.java
jcantrill: @adietish add a boolean  to determine if updated?  We can. I'm not opposed.  What's the value?

src/main/java/com/openshift/internal/restclient/capability/resources/ProjectTemplateListCapability.java
adietish: I was just trying to grasp the semantics of #isSupported() (aka "what does this semantically mean if you get a <code>true</code> from such a call, when would it fail etc.

src/main/java/com/openshift/internal/restclient/capability/resources/ProjectTemplateListCapability.java
adietish: hmm, I still believe that being unauthorized or ony other exception is pretty much a game-killer, The caller would still be able to catch it wherever he wants since we use runtime exceptions.

src/main/java/com/openshift/internal/restclient/model/template/Template.java
adietish: not sure a boolean is enough. Do we provide a means to the user so that he can find out what values were not substituted? What about returning a collection of unknown keys? if it's empty we're all fine, if not it would contain the non-substitued keys (and we could warn the user about these)?

src/main/java/com/openshift/internal/restclient/capability/resources/ProjectTemplateListCapability.java
jcantrill: For this particular case, it should really never fail since you should always be able to list templates.  In theory, I guess, policy could be setup to restrict this operation but I don't know that we can determine that (at least for now) until we try to actually list templates.

src/main/java/com/openshift/internal/restclient/model/template/Template.java
jcantrill: Per our discussion, we will update to a collection of unsub'ed values if needed

src/main/java/com/openshift/internal/restclient/capability/resources/ProjectTemplateListCapability.java
jcantrill: per our discussion, removing call to caller so they can handle as desired.

jcantrill:
@adietish merged

PR #12, opened 2015-04-30
Retrieve templates from common namespace

jcantrill:


jcantrill:
@adietish thought this was in master but was not

PR #13, opened 2015-05-04
[OSJC-161] Add properties to deployment config and fix source of templat...

jcantrill:
...e labels


jcantrill:
@adietish merging to support create application from template

PR #14, opened 2015-05-06
[OSJC-162] Add capability for a resource to expose its client

jcantrill:


jcantrill:
@adietish merged to support 'delete resource'

PR #15, opened 2015-05-08
[OSJC-163] add support for v1beta3 Kubernetes API and OpenShift

jcantrill:


jcantrill:
@adietish Heads up.  Merged v1beta3 api models.  There is an associated minor change to jbosstools but its an unavoidable interface change.  We wont fully support v1beta1, but its upstream's intention to drop support completely of v1beta1 on the server.

PR #18, opened 2015-05-26
[OSJC-169] changed KubernetesResource#equals #hashCode to constant behav

adietish:


src/main/java/com/openshift/internal/restclient/model/KubernetesResource.java
```diff
@@ -105,11 +105,10 @@ public void refresh(){
 		this.node = ModelNode.fromJSONString(client.get(getKind(), getName(), getNamespace()).toString());
 	}
 	
-	// TODO Pretty certain this should be protected
 	@Override
 	public ResourceKind getKind(){
-		if(node.has("kind")){
-			return ResourceKind.valueOf(node.get("kind").asString());
+		if(node.has(ResourcePropertyKeys.KIND)){
```
jcantrill: @adietish this works for both "labels" and "kind" but technically is in correct. ResourcePropertyKeys is really a key value that should never change to a path defined in ResourcePropertiesRegistry.  This path is versioned and gives as the value at defined by the path.  This class has a number of helper methods that retrieve values based on keys and the direct use of node here is probably legacy from when it was first implemented.  We should consider using the helper methods.

src/main/java/com/openshift/internal/restclient/model/KubernetesResource.java
```diff
@@ -161,20 +158,26 @@ public void addLabel(String key, String value) {
...
 	
 	protected void set(String key, String value){
-		String [] property = propertyKeys.get(key);
-		node.get(property).set(value);
+		node.get(getPath(key)).set(value);
 	}
 	
+	private String[] getPath(String key) {
+		String [] property = propertyKeys.get(key);
+		if (property == null) {
```
jcantrill: @adietish we should consider throwing here/providing an indication that the path doesnt exist.  If the path doesnt exist for a given property then it needs to be added to the map IMO.

adietish: @jcantrill isnt the property map a mapping so that if a special mapping is missing it indicates that the key is to be used as is? What is the purpose of these property-keys?

src/main/java/com/openshift/internal/restclient/model/KubernetesResource.java
```diff
@@ -105,11 +105,11 @@ public void refresh(){
...
 	}
 	
-	// TODO Pretty certain this should be protected
 	@Override
 	public ResourceKind getKind(){
-		if(node.has("kind")){
-			return ResourceKind.valueOf(node.get("kind").asString());
+		ModelNode kindNode = get(ResourcePropertyKeys.KIND);
+		if(kindNode.isDefined()){
+			return ResourceKind.valueOf(kindNode.asString());
```
adietish: this will throw for unknown/new kinds that we are not aware of. We should either get away from enums or provide a lenient #valueOf

jcantrill: @adietish this was one of my thoughts to  KubernetesResource not being abstract for the cases of when we do have kinds we dont understand about and we decide to move away from enum

jcantrill:
@adietish LGTM

adietish:
[merge]

PR #20, opened 2015-06-01
Use propertyKeys when setting Labels on KubernetesResource

aslakknutsen:


jcantrill:
@aslakknutsen LGTM. merged

PR #21, opened 2015-06-01
Allow setting IBuildStrategy on BuildConfig

aslakknutsen:
Not sure if this was the idea of the // FIXME, but it in general works.

Any bigger plans beyond this? If not I'll implement similar for BuildSource and BuildTriggers

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/model/BuildConfig.java
+++ b/src/main/java/com/openshift/internal/restclient/model/BuildConfig.java
@@ -9,12 +9,10 @@
 package com.openshift.internal.restclient.model;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
 import org.jboss.dmr.ModelNode;
-import org.jboss.dmr.ModelType;
 
 import com.openshift.internal.restclient.OpenShiftAPIVersion;
 import com.openshift.internal.restclient.model.build.CustomBuildStrategy;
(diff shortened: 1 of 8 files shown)


src/main/java/com/openshift/internal/restclient/model/BuildConfig.java
```diff
@@ -142,6 +200,14 @@ public void setOutput(DockerImageURI imageUri){
 		return null;
 	}
 
+	private void storeEnvironmentVars(final String [] key, ModelNode root, Map<String, String> envs){
```
jcantrill: @aslakknutsen consider making this variable accept a key that will be used to look up a path similiar to the existing convenience "get" methods

aslakknutsen: Right, was wondering about that. I ended up just implementing it as how 'loadEnvrionmentVars' was implemented :)

aslakknutsen: @jcantrill Should be handled in this commit: https://github.com/aslakknutsen/openshift-restclient-java/commit/9b738fb5a8751774f6c450a6026a68ceba057974 

Named them setEnvMap and getEnvMap to differentiate between Kubernetes EnvVar(using name and value) type used here and just  JSON Maps used in e.g. Labels (asMap())

jcantrill:
@aslakknutsen LGTM.  Merged

PR #24, opened 2015-06-01
Allow adding IBuildTrigger's to IBuildConfig

aslakknutsen:


src/main/java/com/openshift/internal/restclient/model/BuildConfig.java
```diff
@@ -77,6 +79,38 @@ public BuildConfig(ModelNode node, IClient client, Map<String, String []> proper
...
+			triggerNode.get(new String[] {"generic", "secret"}).set(generic.getSecret());
+			break;
+		case github:
+			if(!(trigger instanceof IWebhookTrigger)) {
+				throw new IllegalArgumentException("IBuildTrigger of type github does not implement IWebhookTrigger");
+			}
+			IWebhookTrigger github = (IWebhookTrigger)trigger;
+			triggerNode.get(new String[] {"github", "secret"}).set(github.getSecret());
+			break;
+		case imageChange:
```
jcantrill: @aslakknutsen these work for now, but the back-end has moved to object refs which is simpler and only requires you to reference an instance of a kind by name.  Don't quote me, but I think its the name of an imagestream

jcantrill:
@aslakknutsen LGTM.  I think it needs a rebase for me to merge

aslakknutsen:
@jcantrill Branch rebased.

src/main/java/com/openshift/internal/restclient/model/BuildConfig.java
aslakknutsen: @jcantrill Right, just mapping what was there. I'll look at a possible v1beta4? update.

jcantrill:
cc @aslakknutsen  merged

PR #26, opened 2015-06-03
Support for K8 Secret object

jpechane:


src/main/java/com/openshift/internal/restclient/model/Secret.java
```diff
@@ -0,0 +1,72 @@
...
+			addData(key, os.toByteArray());
+		}
+		catch (Exception e) {
+			throw new IllegalArgumentException("Could not process data stream", e);
+		}
+	}
+
+	@Override
+	public void addData(final String key, final byte[] data) {
+		ModelNode dataNode = get(SECRET_DATA);
```
jcantrill: This should never be null as DMR always returns something

src/main/java/com/openshift/restclient/model/secret/ISecret.java
```diff
@@ -0,0 +1,54 @@
...
+	 * @param key
+	 * @return byte representation of the stored data
+	 */
+	byte[] getData(String key);
+	
+	/**
+	 * Sets the type of Secrete - default Opaque
+	 * @param type
+	 */
+	void setType(SecretType type);
```
jcantrill: In future, we will most likely be moving away from proper java enums in favor of something more lenient

src/main/java/com/openshift/internal/restclient/model/Secret.java
```diff
@@ -0,0 +1,72 @@
...
+import com.openshift.restclient.utils.Base64Coder;
+
+/**
+ * @author Jiri Pechanec
+ */
+public class Secret extends KubernetesResource implements ISecret {
+
+	public Secret(ModelNode node, IClient client, Map<String, String []> propertyKeys) {
+		super(node, client, propertyKeys);
+		initializeCapabilities(getModifiableCapabilities(), this, getClient());
```
jcantrill: This is not required unless you are initializing specific capabilites that apply to a Secret kind

PR #36, opened 2015-06-19
[JBIDE-20043] Fix equals after enum refactoring where kind was not evaluating correctly

jcantrill:


jcantrill:
@adietish merged to master

PR #37, opened 2015-06-23
Add support for K8 serviceAccount

dsimansk:


src/main/java/com/openshift/restclient/ResourceKind.java
```diff
@@ -98,6 +99,7 @@
 		list.add(RESOURCE_QUOTA);
 		list.add(SERVICE);
 		list.add(SECRET);
+		list.add(SERVICE_ACCOUNT);
```
adietish: @jcantrill to me this list very much like it should be a set

jcantrill: You are correct, this should be a set.  Please feel free to change as needed.  The only place this 'list' is used is to validate the registered types have the proper constructor used by reflection in the ResourceFactory.

----- Original Message -----

src/main/java/com/openshift/internal/restclient/model/ServiceAccount.java
```diff
@@ -0,0 +1,64 @@
...
+		for (ModelNode entry : node.asList()) {
+			secrets.add(entry.get("name").asString());
+		}
+		return secrets;
+	}
+
+	@Override
+	public void addSecret(String secret) {
+		ModelNode secretNode = get(SERVICE_ACCOUNT_SECRETS);
+		secretNode.add("name", secret);
```
jcantrill: There is a class called JbossDmrExtensions that has a number of getter and setter helpers.  I would suggest using them as things like trying to set a null value on a node throws.  Alternative is to do null checks

jcantrill:
@dsimansk please add a testcase for the v1beta3 version since it is still supported at the moment.  I imagine your test will be a copy and paste

src/main/java/com/openshift/internal/restclient/model/ServiceAccount.java
dsimansk: Hi @jcantrill,
I've check `JBossDmrExtensions` and `KubernetesResource#set()` methods. Every variant uses combination of `get()` and `set()` on ModelNode object to retriev entry and change value, but I'd ideally need `add()` method to new pair of secret `name: value` Is it sufficient to add null checks or do you have any other suggestion how to proceed?

jcantrill: On Monday, September 7, 2015, David Simansky <notifications@github.com
<javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:

> In
> src/main/java/com/openshift/internal/restclient/model/ServiceAccount.java
> <https://github.com/openshift/openshift-restclient-java/pull/37#discussion_r38860377>
> :
>
> > +
> > +	@Override
> > +	public Collection<String> getSecrets() {
> > +		Collection<String> secrets = new ArrayList<>();
> > +		ModelNode node = get(SERVICE_ACCOUNT_SECRETS);
> > +		if(node.getType() != ModelType.LIST) return secrets;
> > +		for (ModelNode entry : node.asList()) {
> > +			secrets.add(entry.get("name").asString());
> > +		}
> > +		return secrets;
> > +	}
> > +
> > +	@Override
> > +	public void addSecret(String secret) {
> > +		ModelNode secretNode = get(SERVICE_ACCOUNT_SECRETS);
> > +		secretNode.add("name", secret);
>
> Hi @jcantrill <https://github.com/jcantrill>,
> I've check JBossDmrExtensions and KubernetesResource#set() methods. Every
> variant uses combination of get() and set() on ModelNode object to
> retriev entry and change value, but I'd ideally need add() method to new
> pair of secret name: value Is it sufficient to add null checks or do you
> have any other suggestion how to proceed?
>
Dmr Ali is such that get will always return something even when that value
does not exist. I thought add was more for appending to a list but I would
defer to looking at the api.  Generally, the set methods are written to
retention silently when the value is null

> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openshift/openshift-restclient-java/pull/37/files#r38860377>
> .
>


-- 
--
Jeff Cantrill
Red Hat, Inc.
jcantril@redhat.com
http://www.redhat.com

dsimansk:
Hi, @jcantrill, could you take a look please? I hope finally it's ok. I've added v1beta3 testcase and also test for setting null value.

jcantrill:
[merge]

PR #39, opened 2015-06-24
changed ResourceKind#values to set, smaller cleanups

adietish:


adietish:
@jcantrill  please review

src/test/java/com/openshift/internal/restclient/model/build/WebhookTriggerTest.java
```diff
@@ -29,13 +29,13 @@ public void setup(){
 	}
 	@Test
 	public void testGetWebhookUrlWhenResourceHasBaseURL() {
-		trigger = new WebhookTrigger(BuildTriggerType.generic, "secret101","foo","https://localhost:8443","v1beta1","test");
-		assertEquals("https://localhost:8443/osapi/v1beta1/buildConfigHooks/foo/secret101/generic?namespace=test", trigger.getWebhookURL());
+		trigger = new WebhookTrigger(BuildTriggerType.GENERIC, "secret101","foo","https://localhost:8443","v1beta1","test");
+		assertEquals("https://localhost:8443/osapi/v1beta1/buildConfigHooks/foo/secret101/Generic?namespace=test", trigger.getWebhookURL());
```
adietish: ...et101/Generic?
             ^^^^^^^^^^ 
Is this correct? do we relly have a capital "G" in this url?

jcantrill: Hmm...we do have some mixed cased URL values though since v1beta3, I believe case no longer matters.  Recall this enum is used in the API models where it is more significant then in the URL string.

src/main/java/com/openshift/internal/restclient/ResourceFactory.java
```diff
@@ -104,7 +104,7 @@
 	}
 	private IClient client;
 	
-	public ResourceFactory(IClient client) {
+	protected ResourceFactory(IClient client) {
```
jcantrill: Why are we changing the visibility?  It's labeled an internal package and if we are really concerned about extension maybe we should just mark it as final

adietish: I was thinking that we wouldnt want anyone outside to be able to instantiate the factory. Final would be something different, it was disallow subclassing. Dont agree?

src/test/java/com/openshift/internal/restclient/model/build/WebhookTriggerTest.java
adietish: I just came across this when I was replacing the deprecated enum in the test and noticed the test failing because of the changed casing. Was thus wondering if the result was right.

jcantrill: equalsIgnoreCase would be fine

src/main/java/com/openshift/internal/restclient/ResourceFactory.java
jcantrill: We can change it back if we find it causes some difficulties

src/test/java/com/openshift/internal/restclient/model/build/WebhookTriggerTest.java
adietish: Imho an exact expected match is what we want to assert, since the logic, that casing wouldnt matter is in the server and we dont want to assert this, right? Thus to me having the exact URL with uppercase in "Generic" is what I want to assert.

adietish:
merged and pushed to upstream/master

PR #51, opened 2015-07-26
[OSJC-197] Provide delay to allow time for the process to start before checking if it has exited

jcantrill:


jcantrill:
[test]

jcantrill:
[merge]

src/main/java/com/openshift/internal/restclient/capability/resources/OpenShiftBinaryPortForwarding.java
```diff
@@ -103,9 +103,15 @@ private void startProcess(String location, PortPair [] ports) {
 	}
 	
 	private void checkProcessIsAlive() throws IOException {
-		Thread.yield();
-		if(!process.isAlive() && process.exitValue() != 0) {
-			throw new OpenShiftException("Port forwarding process exited: %s", IOUtils.toString(process.getErrorStream()));
+		try {
+			Thread.sleep(1000);
```
adietish: @jcantrill I doubt that a fixed timeout is the right thing to do

PR #63, opened 2015-08-26
Add error param handling for OAuth redirects

liggitt:


liggitt:
@jcantrill PTAL

jcantrill:
LGTM.  we have the bot hooked up here for both merge and test

liggitt:
I'm getting closer to being able to test

liggitt:
tested, updated to call `createLocationURI` instead of `getLocationURI`, because calling `getLocationURI` twice with the same URI raises a redirect loop exception. If we called it to check stuff, then called the `super` implementation, it would also call `getLocationURI` and think it had hit a redirect loop, rather than following the URI.

liggitt:
[test]

liggitt:
@jcantrill all set, PTAL

liggitt:
[merge]

jcantrill:
[merge]

PR #69, opened 2015-09-09
Override getName method

tomason:
If this is not done and there is an error creating the resource on server, NPE is thrown.


jcantrill:
[test]

jcantrill:
[merge]

PR #74, opened 2015-10-14
JBIDE-20763 : store 'oc' binary in ThreadLocal

fbricon:
Signed-off-by: Fred Bricon <fbricon@gmail.com>

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/capability/resources/AbstractOpenShiftBinaryCapability.java
+++ b/src/main/java/com/openshift/internal/restclient/capability/resources/AbstractOpenShiftBinaryCapability.java
@@ -18,6 +18,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.openshift.restclient.IClient;
+import com.openshift.restclient.OpenShiftContext;
 import com.openshift.restclient.OpenShiftException;
 import com.openshift.restclient.authorization.BasicAuthorizationStrategy;
 import com.openshift.restclient.authorization.IAuthorizationStrategyVisitor;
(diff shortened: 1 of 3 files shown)


src/main/java/com/openshift/restclient/OpenShiftBinary.java
```diff
@@ -0,0 +1,121 @@
...
+	 * Returns the {@link OpenShiftBinary} instance for the current {@link Thread}. 
+	 * If no instance was set via the {@link #setInstance(OpenShiftBinary)} method, 
+	 * an attempt is made to initialize a new instance from the 
+	 * "openshift.restclient.oc.location" system property.
+	 * 
+	 * @return an {@link OpenShiftBinary}
+	 * @throws LocationNotFoundException if no {@link OpenShiftBinary} can be found
+	 */
+	public static OpenShiftBinary getInstance() {
+		return threadLocalBinary.get();
```
adietish: why use a thread local if a simple static field would work?

fbricon:
The idea is to be able to use different binaries (per OS connection), as we've seen incompatibilities already between executables and OS platforms already. Using a threadlocal makes it more thread safe than having one global static variable.

jcantrill:
@fbricon  so i dont like the fact that now the consumer is dependent upon  a class called OpenShiftBinary; it kind of defeats the purpose for the interface abstraction where we could swap it out for a pure java impl in the future.  I suggest we move OpenShiftBinary to some form of 'Context<T>' which would allow it to be less specific and reused.  Additionally we could make the T a map that would allow passing multiple values if needed.

fbricon:
Updated PR to use a Map<String,Object>

jcantrill:
LGTM [merge]

PR #75, opened 2015-10-20
additions for v3 jenkins plugin

gabemontero:
@jcantrill @adietish  @fbricon :  per discussion with @jcantrill , I'm alerting each of you to this PR, which is intended to extend the some of the REST resources to expose information needed by the V3 Jenkins plug-in work I'm doing.

Of course let me know whatever changes/additions you all want done prior to @jcantrill merging and deploying to the maven repo.

thanks

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
+++ b/src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
@@ -8,6 +8,7 @@ import java.util.Map.Entry;
 import java.util.Set;
 
 import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
 
 import com.openshift.internal.restclient.model.deploy.DeploymentTrigger;
 import com.openshift.internal.restclient.model.deploy.ImageChangeTrigger;
@@ -31,10 +32,21 @@ public class DeploymentConfig extends KubernetesResource implements IDeploymentC
 	private static final String DEPLOYMENTCONFIG_TEMPLATE_LABELS = "spec.template.metadata.labels";
 	private static final String DEPLOYMENTCONFIG_TRIGGERS = "spec.triggers";
 	private static final String DEPLOYMENTCONFIG_STRATEGY = "spec.strategy.type";
+	private static final String DEPLOYMENTCONFIG_LATEST_VERSION = "status.latestVersion";
+	private static final String DEPLOYMENTCONFIG_CAUSES = "status.details.causes";
 	
 	private static final String IMAGE = "image";
 	private static final String ENV = "env";
 	private static final String TYPE = "type";
+	private static final String IMAGE_CHANGE = "ImageChange";
+	private static final String IMAGE_TRIGGER = "imageTrigger";
+	private static final String FROM = "from";
+	private static final String NAME = "name";
+	private static final String IMAGE_CHANGE_PARAMS = "imageChangeParams";
+	private static final String LAST_TRIGGER_IMAGE = "lastTriggeredImage";
+	private static final String IMAGE_TRIGGER_FROM_NAME =  IMAGE_TRIGGER + "." + FROM + "." + NAME;
+	private static final String IMAGE_CHANGE_PARAMS_FROM_NAME = IMAGE_CHANGE_PARAMS + "." + FROM + "." + NAME;
+	private static final String IMAGE_CHANGE_PARAMS_LAST_TRIGGER_IMAGE = IMAGE_CHANGE_PARAMS + "." + LAST_TRIGGER_IMAGE;
 
 	private final Map<String, String[]> propertyKeys;
 
@@ -81,6 +93,20 @@ public class DeploymentConfig extends KubernetesResource implements IDeploymentC
 		return names;
 	}
 	
+	
+	
+	@Override
+	public int getLatestVersionNumber() {
+		return asInt(DEPLOYMENTCONFIG_LATEST_VERSION);
+	}
+	
+	
+
+	@Override
+	public void setLatestVersionNumber(int newVersionNumber) {
+		set(DEPLOYMENTCONFIG_LATEST_VERSION, newVersionNumber);
+	}
+
 	@Override
 	public int getReplicas(){
 		return asInt(DEPLOYMENTCONFIG_REPLICAS);
@@ -164,4 +190,67 @@ public class DeploymentConfig extends KubernetesResource implements IDeploymentC
 	public String getDeploymentStrategyType() {
 		return asString(DEPLOYMENTCONFIG_STRATEGY);
 	}
+
+	@Override
+	public boolean haveTriggersFired() {
+		ModelNode causes = get(DEPLOYMENTCONFIG_CAUSES);
+		if (causes.getType() == ModelType.UNDEFINED || causes.getType() != ModelType.LIST)
+			return false;
+		return causes.asList().size() > 0;
+	}
+
+	@Override
+	public boolean didImageTrigger(String imageNameTag) {
+		if (!haveTriggersFired() || imageNameTag == null)
+			return false;
+		ModelNode causes = get(DEPLOYMENTCONFIG_CAUSES);
+		if (causes.getType() == ModelType.UNDEFINED ||
+			causes.getType() != ModelType.LIST)
+			return false;
+		for (ModelNode cause : causes.asList()) {
+			String type = asString(cause, TYPE);
+			if (type.equalsIgnoreCase(IMAGE_CHANGE)) {
+				String triggerName = asString(cause, IMAGE_TRIGGER_FROM_NAME);
+				DockerImageURI uri = new DockerImageURI(triggerName);
+				if (imageNameTag.equals(uri.getNameAndTag()))
+					return true;
+			}
+		}
+		return false;
+	}
+
+	@Override
+	public String getImageHexIDForImageNameAndTag(String imageNameTag) {
+		ModelNode triggers = get(DEPLOYMENTCONFIG_TRIGGERS);
+		if (triggers.getType() == ModelType.UNDEFINED || triggers.getType() != ModelType.LIST || imageNameTag == null)
+			return null;
+		for (ModelNode trigger : triggers.asList()) {
+			if (asString(trigger, TYPE).equalsIgnoreCase(IMAGE_CHANGE)) {
+				String nameTag = null;
+				nameTag = asString(trigger, IMAGE_CHANGE_PARAMS_FROM_NAME);
+				if (imageNameTag.equals(nameTag)) {
+					return asString(trigger, IMAGE_CHANGE_PARAMS_LAST_TRIGGER_IMAGE);
+				}
+			}
+		}
+		return null;
+	}
+
+	@Override
+	public String getImageNameAndTagForTriggeredDeployment() {
+		ModelNode causes = get(DEPLOYMENTCONFIG_CAUSES);
+		if (causes.getType() == ModelType.UNDEFINED || causes.getType() != ModelType.LIST)
+			return null;
+		for (ModelNode cause : causes.asList()) {
+			String type = asString(cause, TYPE);
+			if (type.equalsIgnoreCase(IMAGE_CHANGE)) {
+				String imageTag = asString(cause, IMAGE_TRIGGER_FROM_NAME);
+				DockerImageURI uri = new DockerImageURI(imageTag);
+				return uri.getNameAndTag();
+			}
+		}
+		return null;
+	}
+	
+	
 }
--- a/src/main/java/com/openshift/internal/restclient/model/ImageStream.java
+++ b/src/main/java/com/openshift/internal/restclient/model/ImageStream.java
@@ -8,9 +8,11 @@
  ******************************************************************************/
 package com.openshift.internal.restclient.model;
 
+import java.util.List;
 import java.util.Map;
 
 import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
 
 import com.openshift.restclient.IClient;
 import com.openshift.restclient.images.DockerImageURI;
@@ -22,6 +24,11 @@ import com.openshift.restclient.model.IImageStream;
 public class ImageStream extends KubernetesResource implements IImageStream {
 
 	private static final String IMAGESTREAM_DOCKER_IMAGE_REPO = "spec.dockerImageRepository";
+	private static final String IMAGESTREAM_SPEC_TAGS = "spec.tags";
+	private static final String IMAGESTREAM_STATUS_TAGS = "status.tags";
+	private static final String TAG = "tag";
+	private static final String ITEMS = "items";
+	private static final String IMAGE = "image";
 
 	public ImageStream(){
 		this(new ModelNode(), null, null);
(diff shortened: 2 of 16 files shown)


src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
```diff
@@ -164,4 +188,79 @@ public IDeploymentTrigger addTrigger(String type) {
 	public String getDeploymentStrategyType() {
 		return asString(DEPLOYMENTCONFIG_STRATEGY);
 	}
+
+	@Override
+	public boolean haveTriggersFired() {
+		ModelNode causes = get(DEPLOYMENTCONFIG_CAUSES);
+		if (causes == null || causes.getType() != ModelType.LIST)
```
jcantrill: The underlying impl is DMR which never returns null; it always returns something.  'null' is not possible but ModelType.UNDEFINED is.

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
```diff
@@ -164,4 +188,79 @@ public IDeploymentTrigger addTrigger(String type) {
...
+		return causes.asList().size() > 0;
+	}
+
+	@Override
+	public boolean didImageTrigger(String imageTag) {
+		if (!haveTriggersFired())
+			return false;
+		ModelNode causes = get(DEPLOYMENTCONFIG_CAUSES);
+		for (ModelNode cause : causes.asList()) {
+			String type = cause.get(TYPE).asString();
```
jcantrill: The base class has 'asString(ModelNode,key)' to do this transformation.  Additionally the 'asXXXXX' methods return defaults instead of 'undefined' nodes

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
```diff
@@ -164,4 +188,79 @@ public IDeploymentTrigger addTrigger(String type) {
...
+			if (type.equalsIgnoreCase(IMAGE_CHANGE)) {
+				String triggerName = cause.get(IMAGE_TRIGGER).get(FROM).get(NAME).asString();
```
jcantrill: Does triggername conform to a docker uri (e.g. repo/namespace/image:tag)? If it does there is a class called 'DockerURI' that will parse out the segments as needed

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
```diff
@@ -164,4 +188,79 @@ public IDeploymentTrigger addTrigger(String type) {
...
+				}
+			}
+		}
+		return false;
+	}
+
+	@Override
+	public String getImageHexIDForImageTag(String imageTag) {
+		ModelNode triggers = get(DEPLOYMENTCONFIG_TRIGGERS);
+		if (triggers == null || triggers.getType() != ModelType.LIST)
```
jcantrill: Same as previous.  never null but ModelType.UNKOWN

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
```diff
@@ -164,4 +188,79 @@ public IDeploymentTrigger addTrigger(String type) {
...
+			return null;
```
jcantrill: Should we return empty string instead of null?

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
```diff
@@ -164,4 +188,79 @@ public IDeploymentTrigger addTrigger(String type) {
...
+		for (ModelNode trigger : triggers.asList()) {
+			if (trigger.get(TYPE).asString().equalsIgnoreCase(IMAGE_CHANGE)) {
```
jcantrill: asString()

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
```diff
@@ -164,4 +188,79 @@ public IDeploymentTrigger addTrigger(String type) {
...
+				String tag = null;
+				tag = trigger.get(IMAGE_CHANGE_PARAMS).get(FROM).get(NAME).asString();
```
jcantrill: asString(trigger,"imagechangeparams.from.name")

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
```diff
@@ -164,4 +188,79 @@ public IDeploymentTrigger addTrigger(String type) {
...
+				if (imageTag.equals(tag)) {
+					return trigger.get(IMAGE_CHANGE_PARAMS).get(LAST_TRIGGER_IMAGE).asString();
```
jcantrill: asString(trigger,imagechangeparams.lasttriggerimage)

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
```diff
@@ -164,4 +188,79 @@ public IDeploymentTrigger addTrigger(String type) {
...
+				}
+			}
+		}
+		return null;
```
jcantrill: same here? should we return "" instead of null?

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
```diff
@@ -164,4 +188,79 @@ public IDeploymentTrigger addTrigger(String type) {
...
+	@Override
+	public boolean haveTriggersFired() {
+		ModelNode causes = get(DEPLOYMENTCONFIG_CAUSES);
+		if (causes == null || causes.getType() != ModelType.LIST)
+			return false;
+		return causes.asList().size() > 0;
+	}
+
+	@Override
+	public boolean didImageTrigger(String imageTag) {
```
jcantrill: Reviewing this class and the origin api docs, I'm wondering if we should introduce a DeploymentCause class and interface.

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
```diff
@@ -164,4 +188,79 @@ public IDeploymentTrigger addTrigger(String type) {
...
+				if (imageTag.equals(tag)) {
+					return trigger.get(IMAGE_CHANGE_PARAMS).get(LAST_TRIGGER_IMAGE).asString();
+				}
+			}
+		}
+		return null;
+	}
+
+	@Override
+	public String getImageTagForTriggeredDeployment() {
```
jcantrill: Many of the same comments apply to this method as well

src/main/java/com/openshift/internal/restclient/model/ImageStream.java
```diff
@@ -41,4 +48,45 @@ public DockerImageURI getDockerImageRepository() {
 		return new DockerImageURI(asString(IMAGESTREAM_DOCKER_IMAGE_REPO));
 	}
 
+	@Override
+	public void setTag(String newTag, String fromTag) {
+		ModelNode tags = get(IMAGESTREAM_SPEC_TAGS);
+		ModelNode tag = new ModelNode();
+		tag.get("name").set(newTag);
```
jcantrill: 'set' Helper methods on base class will do this for you

src/main/java/com/openshift/internal/restclient/model/ImageStream.java
```diff
@@ -41,4 +48,45 @@ public DockerImageURI getDockerImageRepository() {
...
+		ModelNode from = new ModelNode();
+		from.get("kind").set("ImageStreamTag");
+		from.get("name").set(fromTag);
+		tag.get("from").set(from);
+		tags.add(tag);
+	}
+
+	@Override
+	public String getImageId(String tagName) {
```
jcantrill: Same suggestions as previously

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
```diff
@@ -164,4 +188,79 @@ public IDeploymentTrigger addTrigger(String type) {
...
+	public boolean didImageTrigger(String imageTag) {
+		if (!haveTriggersFired())
+			return false;
+		ModelNode causes = get(DEPLOYMENTCONFIG_CAUSES);
+		for (ModelNode cause : causes.asList()) {
+			String type = cause.get(TYPE).asString();
+			if (type.equalsIgnoreCase(IMAGE_CHANGE)) {
+				String triggerName = cause.get(IMAGE_TRIGGER).get(FROM).get(NAME).asString();
+				if (triggerName.contains("/")) {
+					StringTokenizer st = new StringTokenizer(imageTag, "/");
```
adietish: StringTokenizer is legacy/deprecated, highly inefficient. See [javadoc](http://docs.oracle.com/javase/7/docs/api/java/util/StringTokenizer.html)
> StringTokenizer is a legacy class that is retained for compatibility reasons although
> its use is discouraged in new code. It is recommended that anyone seeking this functionality 
> use the split method of String or the java.util.regex package instead
Please use String.split() instead,

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
adietish: I'd argue for the null since that's usually what you'd use to signal that you couldnt find the value/item

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
```diff
@@ -164,4 +188,79 @@ public IDeploymentTrigger addTrigger(String type) {
...
+		String imageTag = null;
+		ModelNode causes = get(DEPLOYMENTCONFIG_CAUSES);
+		if (causes == null || causes.getType() != ModelType.LIST)
+			return null;
+		for (ModelNode cause : causes.asList()) {
+			String type = cause.get(TYPE).asString();
+			if (type.equalsIgnoreCase(IMAGE_CHANGE)) {
+				imageTag = cause.get(IMAGE_TRIGGER).get(FROM).get(NAME).asString();
+				if (imageTag.contains("/")) {
+					StringTokenizer st = new StringTokenizer(imageTag, "/");
```
adietish: please use String.split() instead

src/main/java/com/openshift/internal/restclient/model/ImageStream.java
```diff
@@ -41,4 +48,45 @@ public DockerImageURI getDockerImageRepository() {
...
+		if (tags == null || tags.getType() != ModelType.LIST)
+			return null;
+		List<ModelNode> tagWrappers = tags.asList();
+		if (tagWrappers != null) {
+			for (ModelNode tagWrapper : tagWrappers) {
+				ModelNode tag = tagWrapper.get(TAG);
+				ModelNode items = tagWrapper.get(ITEMS);
+				if (tag != null && tag.asString().equals(tagName) && items != null) {
+					List<ModelNode> itemWrappers = items.asList();
+					for (ModelNode itemWrapper : itemWrappers) {
```
adietish: this could be extracted to a separate method #getImageId(List<ModelNode> itemWrappers) for added code clarity/brevity

src/main/java/com/openshift/internal/restclient/model/build/BuildRequest.java
```diff
@@ -24,9 +24,35 @@
...
+	
 
 	public BuildRequest(ModelNode node, IClient client, Map<String, String[]> propertyKeys) {
 		super(node, client, propertyKeys);
 	}
 
+
+	@Override
+	public void setCommitId(String commitId) {
+		ModelNode node = this.getNode();
```
jcantrill: Use helper methods here.

jcantrill:
We should also add tests for setters here: https://github.com/openshift/openshift-restclient-java/blob/master/src/test/java/com/openshift/internal/restclient/model/v1/DeploymentConfigTest.java  and logic tests like https://github.com/openshift/openshift-restclient-java/blob/master/src/test/java/com/openshift/internal/restclient/model/ProjectTest.java which we have for tests that are not explicitly releated to impl. of the get/set type methods

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
gabemontero: addressed in next push - thx

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
gabemontero: addressed in next push - thx

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
gabemontero: In fact it does ("172.30.71.32:5000/test/origin-nodejs-sample:latest" for example).  I'll switch to `DockerImageURI` - thx

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
gabemontero: part of next push

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
gabemontero: My thinking was the same as @adietish here

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
gabemontero: part of next push

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
gabemontero: part of next push

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
gabemontero: same response ... but let me know if @adietish and I have not swayed you

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
gabemontero: my initial interpretation of the restclient data model was that they were generally objects at the granularity of items retrievable via `oc get` (though IBuildRequest was an exception which made sense to me).  Deployment causes would not qualify under that rationale.  So my initial reaction was that it would be overkill, but I for sure would not qualify my opinion here as resolute.

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
gabemontero: yep ... part of next push

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
gabemontero: switched to DockerImageURK (which uses split()) per comment from @jcantrill - StringTokenizer gone.

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
gabemontero: switched to DockerImageURI (which uses split()) per comment from @jcantrill - StringTokenizer gone.

src/main/java/com/openshift/internal/restclient/model/ImageStream.java
gabemontero: yep that is better ... part of next push

src/main/java/com/openshift/internal/restclient/model/build/BuildRequest.java
gabemontero: yep - part of next push

src/main/java/com/openshift/internal/restclient/model/DeploymentConfig.java
```diff
@@ -164,4 +190,64 @@ public IDeploymentTrigger addTrigger(String type) {
...
+		if (causes.getType() != ModelType.LIST)
+			return false;
+		return causes.asList().size() > 0;
+	}
+
+	@Override
+	public boolean didImageTrigger(String imageNameTag) {
+		if (!haveTriggersFired() || imageNameTag == null)
+			return false;
+		ModelNode causes = get(DEPLOYMENTCONFIG_CAUSES);
```
jcantrill: You may need to compare against ModelType.UNDEFINED for causes where the node doesnt exist

gabemontero: missed that one ... part of next push

gabemontero:
@jcantrill - as part of our meeting thursday let's screen share and I can show you my first passes at encapsulating via capabilities build cancel and build on specific commit id

gabemontero:
@jcantrill hey, the new/updated capabilities, along with the undefined check, have been pushed.

jcantrill:
[test]

jcantrill:
[merge]

PR #77, opened 2015-10-22
[OSJC-222] Flush out replication controller attributes and methods

jcantrill:
This PR:
* Adds getters and setters to replication controller
* deploymentconfig extends replication controller
* adds IContainer type


jcantrill:
@adietish @fbricon flushed out support for replication controllers

jcantrill:
[merge]

PR #80, opened 2015-10-29
Support for Persistent Volumes

josefkarasek:
This should enable us to create volumes from java.


src/main/java/com/openshift/internal/restclient/model/PersistentVolume.java
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	/**
+	 * Sets the capacity. There is no conversion between units.
+	 * @param capacity the capacity
+	 * @param unit the unit
+	 */
+	@Override
+	public void setCapacity(long capacity, MemoryUnit unit) {
+		get(PV_CAPACITY).set(capacity + unit.name());
```
jcantrill: There are convenience methods as part of KubernetesResource that provide this logic, set(path,value)

src/main/java/com/openshift/internal/restclient/model/PersistentVolume.java
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	/**
+	 * @return access modes
+	 * @see {@link com.openshift.restclient.model.volume.PVCAccessModes}
+	 * @see
+	 */
+	@Override
+	public Set<String> getAccessModes() {
+		Set<String> modes = new HashSet<>();
```
jcantrill: Depending upon when you started the PR, their is also an 'asSet(path,type)' convenience method that might already do this.

src/main/java/com/openshift/restclient/model/volume/IPersistentVolume.java
```diff
@@ -0,0 +1,46 @@
...
+	MemoryUnit getCapacityUnit();
+	void setCapacity(long capacity, MemoryUnit unit);
+
+	Set<String> getAccessModes();
+	void setAccessModes(String... modes);
+
+	String getReclaimPolicy();
+	void setReclaimPolicy(String policy);
+
+	NFSVolume getNFSVolume();
```
jcantrill: I would expect this to return an instance of IVolume and the impl would provide one of these: https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L169.

src/main/java/com/openshift/restclient/model/volume/IPersistentVolume.java
```diff
@@ -0,0 +1,46 @@
...
+	/**
+	 * @see https://docs.openshift.com/enterprise/3.0/rest_api/kubernetes_v1.html#v1-nfsvolumesource
+	 */
+	void setNFSVolume(NFSVolume volume);
+
+	String getHostPathVolume();
```
jcantrill: I would expect this get/set method to be part of IVolume since that is common across storage implementations.

src/main/java/com/openshift/internal/restclient/model/volume/NFSVolume.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Contributors:
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package com.openshift.internal.restclient.model.volume;
+
+import java.util.Objects;
+
+
+public class NFSVolume {
```
jcantrill: Would expect to implement an INFSVolume interface.  We have discussed modifying the backing JSON library and this would mitigate consumers of carrying about this specific implementation

jcantrill:
@josefkarasek Thank you for the contribution.  Generally, it looks good, but please implement abstractions of the volume for the PV kind. Even though we are not trying to mirror the types exactly or what we could get if we generated the code, this is the pattern we have used in other places (e.g. buildtriggers).

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -49,6 +47,9 @@
 import com.openshift.restclient.model.IResource;
 import com.openshift.restclient.model.user.IUser;
 
+import static com.openshift.internal.restclient.capability.CapabilityInitializer.initializeClientCapabilities;
```
jcantrill: Unused import?

src/main/java/com/openshift/internal/restclient/model/PersistentVolume.java
```diff
@@ -0,0 +1,276 @@
...
+ * Copyright (c) 2015 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package com.openshift.internal.restclient.model;
```
jcantrill: Please move to the '*.model.volume' package

src/main/java/com/openshift/internal/restclient/model/PersistentVolume.java
```diff
@@ -0,0 +1,276 @@
...
+
+	/**
+	 * Sets the access modes.
+	 * If there are any modes present, they are overridden by the {@code modes} parameter.
+	 * @param modes
+	 *             the access modes
+	 */
+	@Override
+	public void setAccessModes(String... modes) {
+		ModelNode accessModeNode = get(PV_ACCESS_MODES);
```
jcantrill: Convenience setter for this...though it takes a 'Set' but could a new one could be added to take an array

src/main/java/com/openshift/internal/restclient/model/PersistentVolume.java
```diff
@@ -0,0 +1,276 @@
...
+	/**
+	 * Deletes all PV types.
+	 * Kubernetes configuration doesn't define consistent definition
+	 * of different PV types. Need to check for all possibilities.
+	 *
+	 * @see 'https://github.com/kubernetes/kubernetes/issues/16155'
+	 * @see 'https://github.com/kubernetes/kubernetes/blob/a10ed3e2b178096dd9575874f7e5916ff244dcea/docs/user-guide/volumes.md#types-of-volumes'
+	 *
+	 */
+	private void removeAllPVTypes() {
```
jcantrill: I dont think this will be necessary if impl. a single volume source as defined in the kube types: https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L169  Most of the patterns upstream are to have a 'type' field to tell you what type the actual volume is.  The expectation is that only one kind will be populated while the others are empty.

src/main/java/com/openshift/internal/restclient/model/PersistentVolume.java
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	/**
+	 * @param unit the designated unit. One of "Ki", "Mi", "Gi", "Ti", "Pi", "Ei".
+	 * @return 0, if conversion not possible. Otherwise the capacity in given units.
+	 * @see {@link PersistentVolume#convert(String, MemoryUnit)}
+	 * @throws IllegalArgumentException if not supported {@code unit} given.
+	 */
+	@Override
+	public long getCapacity(String unit) {
```
jcantrill: I don't know the answer for these capacity methods, but do they need to be BigInteger?

josefkarasek: Long certainly can not hold big values in exbibytes. My assumption is that by the time we'll be creating single volumes with capacity of several Ei's, OSE3 will have been deprecated. I find it to be more convenient to use a primitive type, since it fits the most cases (if not all) and is easier to work with.

src/main/java/com/openshift/restclient/model/volume/IPersistentVolume.java
josefkarasek: HostPath is actually one of kubernetes resources. I will add common method getPath to IVolume and rewrite getVolume to be more generic.

src/main/java/com/openshift/internal/restclient/DefaultClient.java
josefkarasek: That's probably caused by my IDE setting. Didn't notice that, sorry.

src/main/java/com/openshift/restclient/model/volume/property/IPersistentVolumeProperties.java
```diff
@@ -0,0 +1,10 @@
+package com.openshift.restclient.model.volume.property;
+
+
+import org.jboss.dmr.ModelNode;
+
+public interface IPersistentVolumeProperties {
+
+	void setProperties(ModelNode node);
```
josefkarasek: Exposing implementation detail in public interface - workaround for the fact, that the format of type properties can generally differ.
This way I can avoid deciding what type of properties the client code sends, otherwise there would have to be either instanceof or isAssignableFrom to check type of the object it received.

src/main/java/com/openshift/restclient/model/volume/property/IPersistentVolumeProperties.java
```diff
@@ -0,0 +1,10 @@
+package com.openshift.restclient.model.volume.property;
+
+
+import org.jboss.dmr.ModelNode;
+
+public interface IPersistentVolumeProperties {
```
josefkarasek: I decided to go with entirely new interface instead of using IVolume, because PersistentVolume is type of Volume as a whole. But persistent volume Type are just property values for PersistentVolume.

jcantrill: WFM

src/main/java/com/openshift/restclient/model/volume/property/IPersistentVolumeProperties.java
jcantrill: We dont want to expose ModelNode on the the interface.  Based on how is consumed,  this is really a copy(IPersistentVolumeProperties) or clone(IPersistentVolumeProperties) method.  I am not fond of casting or type checking either, but please impl a  common base class to the impls (e.g. ModelNodeAdapter)  which would allow minimal casting and type checking to accomplish the same thing without exposing ModelNode

jcantrill:
[test]

jcantrill:
[merge]

PR #81, opened 2015-10-29
make IParameters clonable

jcantrill:
[merge]

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/model/template/Parameter.java
+++ b/src/main/java/com/openshift/internal/restclient/model/template/Parameter.java
@@ -30,6 +30,12 @@ public class Parameter implements IParameter{
 		this.node = node;
 	}
 	
+	
+	@Override
+	public IParameter clone() {
+		return new Parameter(node.clone());
+	}
+
 	@Override
 	public String getName() {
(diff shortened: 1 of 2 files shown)


src/main/java/com/openshift/internal/restclient/model/template/Parameter.java
```diff
@@ -30,6 +30,12 @@ public Parameter(ModelNode node){
 		this.node = node;
 	}
 	
+	
+	@Override
+	public IParameter clone() {
```
fbricon: this needs to be declared on the interface. You can't call ITemplate.clone() as ITemplate is not an Object

current workaround is to call commons lang ObjectUtils.clone(object), which uses reflection to call clone()

src/main/java/com/openshift/restclient/model/template/IParameter.java
```diff
@@ -13,7 +13,7 @@
  * 
  * @author Jeff Cantrill
  */
-public interface IParameter {
+public interface IParameter extends Cloneable {
```
fbricon: Cloneable doesn't declare a clone() method

PR #83, opened 2015-10-30
[OSJC-225] Add watch support for resources

jcantrill:
@fbricon @adietish Please review

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/WatchClient.java
+++ b/src/main/java/com/openshift/internal/restclient/WatchClient.java
@@ -1,0 +1,199 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package com.openshift.internal.restclient;
+
+import java.io.IOException;
+import java.net.URL;
+import java.nio.CharBuffer;
+import java.security.KeyManagementException;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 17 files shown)


src/main/java/com/openshift/internal/restclient/WatchClient.java
```diff
@@ -0,0 +1,199 @@
...
+			final String resourceVersion = getResourceVersion(kind, namespace, token);
+			try(CloseableHttpAsyncClient httpClient = HttpAsyncClients
+					.custom()
+					.setSSLContext(sslContext)
+					.setHostnameVerifier(sslCertAdapter).build()){
+				final String endpoint = new URLBuilder(baseUrl, typeMappings)
+						.kind(kind)
+						.namespace(namespace)
+						.watch()
+						.addParmeter("resourceVersion", resourceVersion)
```
fbricon: is addParmeter fixable without breaking everything?

src/main/java/com/openshift/internal/restclient/WatchClient.java
```diff
@@ -0,0 +1,199 @@
...
+			String response = client.execute(get, new BasicResponseHandler());
+			LOGGER.debug(String.format("List Response: %s:", response));
+			IResource list = factory.create(response);
+			return list.getMetadata().get("resourceVersion");
+		}
+	}
+	
+	private HttpUriRequest buildRequest(String endpoint, String token) {
+		HttpGet request = new HttpGet(endpoint);
+		request.addHeader("User-Agent", "openshift-resetclient-java");
```
fbricon: openshift-restclient-java

src/main/java/com/openshift/internal/restclient/WatchClient.java
jcantrill: Not exactly sure what you mean here.

jcantrill:
[merge]

src/main/java/com/openshift/internal/restclient/WatchClient.java
fbricon: typo in the method name addParmeter instead of addParameter. if it's a public method, then renaming it would be problematic for backward compat.

PR #95, opened 2015-12-01
[ OSJC-230] Add volume sources to dc when containers have volume mounts

jcantrill:
This PR:
* Add an emptydir mount source if the container has volumemounts

/cc @adietish @fbricon

### DIFF
--- a/src/main/java/com/openshift/restclient/model/IReplicationController.java
+++ b/src/main/java/com/openshift/restclient/model/IReplicationController.java
@@ -14,6 +14,7 @@ import java.util.Map;
 import java.util.Set;
 
 import com.openshift.restclient.images.DockerImageURI;
+import com.openshift.restclient.model.volume.IVolumeSource;
 
 /**
  * @author Jeff Cantrill
@@ -70,7 +71,7 @@ public interface IReplicationController  extends IResource{
 	 * @param containerPorts  the container ports
 	 * @param volumes		 the set of emptyDir volumes to add to the config
 	 */
-	void addContainer(String name, DockerImageURI tag,  Set<IPort> containerPorts, Map<String, String> envVars, List<String> volumes);
+	IContainer addContainer(String name, DockerImageURI tag,  Set<IPort> containerPorts, Map<String, String> envVars, List<String> volumes);
 
 	/**
 	 * Add a container to the pod that will be spun up as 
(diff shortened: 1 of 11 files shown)


src/main/java/com/openshift/restclient/model/IReplicationController.java
```diff
@@ -70,7 +71,7 @@
 	 * @param containerPorts  the container ports
 	 * @param volumes		 the set of emptyDir volumes to add to the config
 	 */
-	void addContainer(String name, DockerImageURI tag,  Set<IPort> containerPorts, Map<String, String> envVars, List<String> volumes);
+	IContainer addContainer(String name, DockerImageURI tag,  Set<IPort> containerPorts, Map<String, String> envVars, List<String> volumes);
```
fbricon: This breaks the API backward-compat. If this interface was added in 3.1, then it's fine, else, you should consider upversionning the rest client.
Also, it's probably a good idea to start adding @since tags in the javadoc

src/main/java/com/openshift/restclient/model/volume/IVolumeMount.java
```diff
@@ -0,0 +1,21 @@
...
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package com.openshift.restclient.model.volume;
+
+/**
+ * The volumes that are mounted 
+ * in a container
+ * @author jeff.cantrill
+ *
+ */
```
fbricon: you shouldn't extend a deprecated class. Maybe make IVolume extend IVolumeMount instead

jcantrill: Extended the deprecated class with the thought of removing it all together to the correct concept.  I'm hoping to refactor to remove the dependency

src/main/java/com/openshift/restclient/model/IReplicationController.java
jcantrill: Does this really break the backwards compatibility if it adds a return parameter.  User's can still use it without breaking their code.  I'll have to see when it was introduced

jcantrill: Does not appear in the 3.0.x branch so must have been added recently.

fbricon: It doesn't break consumers. It breaks potential implementers. The rest client is the obvious only implementation, but that might break someone's unit tests that mocked that interface.

adietish: 3.1 was released as SNAPSHOT only though, so I would agree in saying, let's do it

src/main/java/com/openshift/restclient/model/volume/IVolumeMount.java
jcantrill: Resolved

jcantrill:
@fbricon refactored to make interface not extend deprecated one

jcantrill:
[merge]

PR #96, opened 2015-12-01
Make default jar output location configurable

fbricon:
With the new jar.outputDir property, one can now automatically generate the rest client jar
directly in the Eclipse org.jboss.tools.openshift.client/lib folder

- Right-click on the openshift-rest-client-java project
- Run As > Maven Build
- set goals=package
- add jar.outpuDir property, set the absolute path to org.jboss.tools.openshift.client/lib as value
- In the refresh tab, enable refresh resources upon completion
- click specify resources..., select org.jboss.tools.openshift.client/lib

Run, profit!

Signed-off-by: Fred Bricon <fbricon@gmail.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -42,7 +42,7 @@
 		
 		<!-- integration test properties file -->
 		<integrationtest.properties>${basedir}/src/test/resources/integrationTest.properties</integrationtest.properties>
-
+		<jar.outputDir>${project.build.directory}</jar.outputDir>
 	</properties>
 
 	<!-- Issues -->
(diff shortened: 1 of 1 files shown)


adietish:
cool! nice addition, thanks

jcantrill:
@fbricon can you add these instructions to the README or a dev guide or something so they are not lost. I will otherwise not remember what they are.

fbricon:
@jcantrill see https://github.com/jbosstools/jbosstools-openshift/blob/master/README.md#working-with-openshift-restclient-java

PR #97, opened 2015-12-09
Sanitize JSON output to remove nulls

jcantrill:
[test]

cc @fbricon @adietish

### DIFF
--- a/src/main/java/com/openshift/internal/util/JBossDmrExtentions.java
+++ b/src/main/java/com/openshift/internal/util/JBossDmrExtentions.java
@@ -8,9 +8,14 @@
  ******************************************************************************/
 package com.openshift.internal.util;
 
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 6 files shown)


src/main/java/com/openshift/internal/util/JBossDmrExtentions.java
```diff
@@ -29,6 +34,59 @@
...
+		if(node.getType() == ModelType.OBJECT) {
+			Collection<String> emptyKeys = new ArrayList<>(node.keys().size());
+			for (String key : node.keys()) {
+				ModelNode child = node.get(key);
+				if(child.getType() == ModelType.UNDEFINED) {
+					node.remove(key);
+				}else {
+					sanitize(child);
+				}
+				if(child.getType() == ModelType.OBJECT && child.keys().size() == 0) {
```
fbricon: do you ever reach that statement?

fbricon:
LTGM but Travis seems to disagree.

My only concern is if we sanitize genuine and significant null values from the server. Better be sure it doesn't bite us in the ass

jcantrill:
[test]

jcantrill:
[merge]

PR #104, opened 2016-01-13
[OSJC-235] DefaultClient equals even if token differs

adietish:


adietish:
@jcantrill plz review

adietish:
[test]

src/main/java/com/openshift/restclient/authorization/TokenAuthorizationStrategy.java
```diff
@@ -38,5 +38,19 @@ public void accept(IAuthorizationStrategyVisitor visitor) {
 		visitor.visit(this);
 	}
 
-
+	@Override
+	public int hashCode() {
```
jcantrill: Why is this so complicated and not just either the hashcode of the token or String.hashCode()

adietish: @jcantrill dont want to rely on the token by choice, since I consider 2 authentication strategies with different tokens to be identical

adietish:
updated the approach:
the DefautClient now wont compare tokens but the strategies. All strategies now report being equal upon comparison of the username (only).
This implies us to update the username in the TokenAuthorizationStrategy once we successfully connected to the server

adietish:
btw. I also:
- dropped the untested/unused KerbrosAuthenticationStrategy
- extracted http constants to IHttpConstants (so that we can drop the deprecated IHttpClient interface in classes that relied on it for those constants only).

src/main/java/com/openshift/restclient/authorization/AbstractAuthorizationStrategy.java
```diff
@@ -0,0 +1,54 @@
...
+	@Override
+	public int hashCode() {
+		final int prime = 31;
+		int result = 1;
+		result = prime * result + ((username == null) ? 0 : username.hashCode());
+		return result;
+	}
+
+	@Override
+	public boolean equals(Object obj) {
```
adietish: now strategies only rely on the username for equality. Kept username unmutable.

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -534,22 +535,17 @@ public boolean equals(Object obj) {
...
-			String token = strategy.getToken();
-			String otherToken = other.strategy == null?null:other.strategy.getToken();
-			if (token == null) {
-				if (otherToken != null)
-					return false;
-			} else if (!token.equals(otherToken))
+			if (other.strategy == null) {
 				return false;
+			}
+			return ObjectUtils.equals(strategy, other.strategy);
```
adietish: DefaultClient#equals wont rely on the token, only on strategy equality.
Strategies only rely on the username for equality (see comment further down).

jcantrill:
LGTM

jcantrill:
[merge]

PR #106, opened 2016-01-20
#105 Allow Route TLS config to be edited

### LINKED ISSUE #105: Cannot modify TLS config of a Route object
TLS config of a route has recently been changed to read only. But the ability to modify this is desired for creating new routes with specific TLS configuration.

[Initial Route.java](https://github.com/openshift/openshift-restclient-java/blob/b51d8745626c28096b20b1956d7879a9621201ee/src/main/java/com/openshift/internal/restclient/model/Route.java#L60)
[Current Route.java](https://github.com/openshift/openshift-restclient-java/blob/ea540325360c23f90456e085b1647580a8480c2e/src/main/java/com/openshift/internal/restclient/model/Route.java#L65)

jcantrill:
fixes #105 by adding a factory method.

cc @adietish @fbricon 

[test]

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/model/Route.java
+++ b/src/main/java/com/openshift/internal/restclient/model/Route.java
@@ -78,6 +78,16 @@ public class Route extends KubernetesResource implements IRoute {
 	}
 	
 	@Override
+	public ITLSConfig createTLSConfig() {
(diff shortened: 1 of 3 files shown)


jcantrill:
[merge]

PR #110, opened 2016-02-03
[JBIDE-21624] Add deploy capability to IDeploymentConfig

jcantrill:
cc @fbricon capability for deploy support in the client.  tools changes will utilize this.
[test]


jcantrill:
[merge]

PR #119, opened 2016-02-29
Add ability to remove annotations

markpaychex:


jcantrill:
[test]

jcantrill:
@markpaychex are you able to use the master branch or do you require this fix in 3.1?  We currently don't have any plans to produce a new 3.1 release.

markpaychex:
Hello,
My assumption was that the master branch (version 4.0.0) is for an upcoming OpenShift version 4.0.0 whereas 3.1.x is for OpenShift 3.1.x. Is that assumption incorrect?

jcantrill:
@markpaychex thats not necessarily true as we bumped to 4.0.0 for a breaking api change needed to support the up coming release of the eclipse tooling.  Openshift will be releasing 3.2 shortly  which will be supported by the 4.x client.  The versioning is confusing I admit, but we definitely should patch the master branch.  If you also need a 3.1 release of the client, we can patch it there as well.

markpaychex:
Ok. In that case it can be put into the master branch I think. Thanks.

jcantrill:
Merged and cherry-picked into master

On Tue, Mar 1, 2016 at 1:27 PM, Mark <notifications@github.com> wrote:

> Ok. In that case it can be put into the master branch I think. Thanks.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openshift/openshift-restclient-java/pull/119#issuecomment-190842729>
> .
>



-- 
--
Jeff Cantrill
Red Hat, Inc.
jcantril@redhat.com
http://www.redhat.com

PR #120, opened 2016-03-02
[OSJC-205] Access raw JSON as object map

jcantrill:
Limited access to underlying json that is impl independent by providing a map.  Also allow access to an individual map entry as a string requiring user to convert.  This is trying to provide some access without fully duplicating DMR

cc @adietish @fbricon

### DIFF
--- a/src/main/java/com/openshift/restclient/capability/resources/IPropertyAccessCapability.java
+++ b/src/main/java/com/openshift/restclient/capability/resources/IPropertyAccessCapability.java
@@ -1,0 +1,65 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package com.openshift.restclient.capability.resources;
+
+import java.util.Map;
+
+import com.openshift.restclient.capability.ICapability;
+
+/**
+ * A mechanism to access the underlying content of a json
+ * structure
+ * 
+ * Relies on a dot delimited path to indentify the root of the entity
+ * being returned (e.g. 'metadata.labels').  This does not provide
+ * a mechanism to retrieve properties across an array kind.
+ * 
+ * @author jeff.cantrill
+ *
+ */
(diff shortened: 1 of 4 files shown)


jcantrill:
[test]

src/main/java/com/openshift/restclient/capability/resources/IPropertyAccessCapability.java
```diff
@@ -0,0 +1,65 @@
...
+	 * The exception thrown when a path given to the capability is
+	 * unresolvable
+	 * 
+	 * @author jeff.cantrill
+	 *
+	 */
+	public static class UnresolvablePathException extends RuntimeException{
+
+		/**
+		 *
```
fbricon: nice javadoc

src/main/java/com/openshift/restclient/capability/resources/IPropertyAccessCapability.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.openshift.restclient.capability.ICapability;
+
+/**
+ * A mechanism to access the underlying content of a json
+ * structure
+ * 
+ * Relies on a dot delimited path to indentify the root of the entity
+ * being returned (e.g. 'metadata.labels').  This does not provide
+ * a mechanism to retrieve properties across an array kind.
```
fbricon: and how bad is it arrays are not supported?

jcantrill: I dont see it as a big deal since you are able to retrieve the entire JSON as a map of string to Objects.  But the bigger question is what does it mean to retrieve an entry across an array?  Do we allow you to do something like "foo.[3].bar" where you retrieve bar value of the third entry?  I think this is probably good enough but if we can come up with a could API, I'm willing to adjust it.

src/main/java/com/openshift/restclient/capability/resources/IPropertyAccessCapability.java
jcantrill: sarcasm?

src/main/java/com/openshift/restclient/capability/resources/IPropertyAccessCapability.java
adietish: I also believe that we should not push this too far since we'd start to get in the area of json parsers which I believe we dont want to do.

jcantrill: I think I agree with @adietish that we want this API as minimal as possible.  It's a last resort to access something that isn't exposed with a proper getter.

jcantrill:
[merge]

PR #124, opened 2016-03-08
[OSJC-244] Retrieve metadata from dockerhub v2 registry

jcantrill:
cc @fbricon @adietish

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/capability/resources/DockerRegistryImageStreamImportCapability.java
+++ b/src/main/java/com/openshift/internal/restclient/capability/resources/DockerRegistryImageStreamImportCapability.java
@@ -1,0 +1,283 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package com.openshift.internal.restclient.capability.resources;
+
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.stream.Collectors;
+
+import org.apache.commons.lang.StringUtils;
+import org.eclipse.jetty.client.HttpClient;
+import org.eclipse.jetty.client.api.ContentResponse;
+import org.eclipse.jetty.client.api.Request;
+import org.eclipse.jetty.http.HttpHeader;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
+import org.jboss.dmr.ModelNode;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.openshift.internal.restclient.model.ModelNodeBuilder;
+import com.openshift.internal.restclient.model.image.ImageStreamImport;
+import com.openshift.internal.restclient.model.properties.ResourcePropertyKeys;
+import com.openshift.internal.util.JBossDmrExtentions;
+import com.openshift.restclient.IResourceFactory;
+import com.openshift.restclient.ResourceKind;
+import com.openshift.restclient.capability.resources.IImageStreamImportCapability;
+import com.openshift.restclient.http.IHttpConstants;
+import com.openshift.restclient.images.DockerImageURI;
+import com.openshift.restclient.model.IProject;
+import com.openshift.restclient.model.image.IImageStreamImport;
+
+/**
+ * Retrieve metadata directly from docker.
+ * @author jeff.cantrill
+ *
+ */
+public class DockerRegistryImageStreamImportCapability implements IImageStreamImportCapability, IHttpConstants, ResourcePropertyKeys {
+	
+	private static final String TOKEN = "token";
+	private static final String STATUS_STATUS = "status.status";
+	private static final int TIMEOUT = 10 * 1000; //10 seconds
+	private static final String ID = "id";
+	private static final String PARENT = "parent";
+	private static final String REALM = "realm";
+	private static final Logger LOG = LoggerFactory.getLogger(IImageStreamImportCapability.class);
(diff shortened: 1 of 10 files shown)


jcantrill:
[test]

jcantrill:
[test]

src/main/java/com/openshift/internal/restclient/capability/resources/DockerRegistryImageStreamImportCapability.java
```diff
@@ -0,0 +1,276 @@
...
+	}
+	
+	@Override
+	public boolean isSupported() {
+		return true;
+	}
+
+	@Override
+	public String getName() {
+		return DockerRegistryImageStreamImportCapability.class.getSimpleName();
```
adietish: you could actually implement this in a base class that then does getClass().getSimpleName().
But since no other common functionality there's not much sense to it right now.

src/main/java/com/openshift/internal/restclient/capability/resources/DockerRegistryImageStreamImportCapability.java
```diff
@@ -0,0 +1,276 @@
...
+	}
+	
+	private boolean registryExists(HttpClient client) throws Exception{
+		ContentResponse response = client.newRequest(DEFAULT_DOCKER_REGISTRY).send();
+		return (response.getStatus() == STATUS_UNAUTHORIZED || response.getStatus() == STATUS_OK);
```
adietish: Can't response be null? ex. if there's something completely different listening and answering?

src/main/java/com/openshift/internal/restclient/capability/resources/DockerRegistryImageStreamImportCapability.java
jcantrill: Agreed.  I don't actually recall why I added it into the original api.

src/main/java/com/openshift/internal/restclient/capability/resources/DockerRegistryImageStreamImportCapability.java
jcantrill: Not that I can ascertain.  It doesn't appear to throw any errors.  As far as I can tell, it wraps up responses and provides an error code, but I could add a null check and simply return false

jcantrill:
[test]

src/main/java/com/openshift/internal/restclient/capability/resources/DockerRegistryImageStreamImportCapability.java
```diff
@@ -0,0 +1,276 @@
...
+	
+	/**
+	 * @return the token required to pull docker metadata
+	 */
+	private String retrieveAuthToken(HttpClient client, String details) throws Exception {
+		if(StringUtils.isNotBlank(details)) {
+			Map<String, String> auth = parseAuthDetails(details);
+			if(auth.containsKey(REALM)) {
+				Request request = client.newRequest(StringUtils.strip(auth.get(REALM),"\""));
+				for (Entry<String, String> e: auth.entrySet()) {
```
adietish: ideal candidate for a method ex. addRealmParameters?

src/main/java/com/openshift/internal/restclient/capability/resources/DockerRegistryImageStreamImportCapability.java
```diff
@@ -0,0 +1,276 @@
...
+	}
+	
+	/**
+	 * @return the token required to pull docker metadata
+	 */
+	private String retrieveAuthToken(HttpClient client, String details) throws Exception {
+		if(StringUtils.isNotBlank(details)) {
+			Map<String, String> auth = parseAuthDetails(details);
+			if(auth.containsKey(REALM)) {
+				Request request = client.newRequest(StringUtils.strip(auth.get(REALM),"\""));
```
adietish: the whole block where you create the request could be a separate method ex. createRequest?

adietish: any idea why we're using the jetty client? is this via websocket? Shouldnt we limit us using the jetty client for websockets and use the apache client in the other cases (or get over to use the jetty client everywhere) - just trying to prevent us from multiplying even more the clients we use.

adietish: is there any specific reason for us to use the jetty client here? I'd rather try to limit us to use the jetty client for websockets and keep sticking to the apache client in all other cases or replace the apache client with the jetty one.

src/main/java/com/openshift/internal/restclient/capability/resources/DockerRegistryImageStreamImportCapability.java
```diff
@@ -0,0 +1,277 @@
...
+			.set("image.dockerImageMetadata", last);
+
+		return buildImageStreamImport(uri, builder);
+	}
+	
+	private ImageStreamImport buildImageStreamImport(DockerImageURI uri, ModelNodeBuilder builder) {
+		ImageStreamImport isImport = (ImageStreamImport) factory.stub(ResourceKind.IMAGE_STREAM_IMPORT, uri.getName(), this.project.getName());
+		ModelNode root = isImport.getNode();
+		ModelNode images = JBossDmrExtentions.get(root,null, ImageStreamImport.STATUS_IMAGES);
+		images.add(builder.build());
```
adietish: any specific reason to build.build() only down here? Why not pass the result of the builder as parameter?

src/main/java/com/openshift/internal/restclient/capability/resources/DockerRegistryImageStreamImportCapability.java
jcantrill: I used the jetty client because it was much more straight forward and obvious how to make this work.  My preference would be to drop the apache client in the other cases we use it if possible because the api really is non-intuative IMO.  We are already using jetty and the addition of the required lib was fairly small.

adietish:
other than those design related comments, +1

src/main/java/com/openshift/internal/restclient/capability/resources/DockerRegistryImageStreamImportCapability.java
jcantrill: fixing

jcantrill:
[merge]

PR #125, opened 2016-03-09
allow import of cert to the trust for ssl validation

gabemontero:
@jcantrill PTAL

This stems from the dialogue with @jim-minter I pulled you into a week or so ago.

This change allows the trust manager in UrlConnectionHttpClient.java to properly validate a server certificate (vs. skipping TLS verification and calling out to the ISSLCertificateCallback).

As an FYI, currently in jenkins-plugin I have to maintain my own TrustStore and validate the cert when UrlConnectionHttpClient calls the ISSLCertificateCallback (btw, this also revealed that the JDK does *NOT* use a singleton pattern when it comes to the TrustStore related stuff ... something I was not expecting :-) ).

Once some form of this change is in the 4.0.0-SNAPSHOT, I'll remove that extra TrustStore in jenkins-plugin and just set the cert being used via ClientBuilder.

Other note:  the trust store in AuthorizationClient was not getting involved in the SSL exchanges in my testing, so I did not update that one.  But let me know if there is some reason you want that one updated as well.

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/http/UrlConnectionHttpClient.java
+++ b/src/main/java/com/openshift/internal/restclient/http/UrlConnectionHttpClient.java
@@ -69,18 +69,64 @@ public class UrlConnectionHttpClient implements IHttpClient {
 	private String excludedSSLCipherRegex;
 	private IAuthorizationStrategy authStrategy;
 
+	private TrustManagerFactory trustManagerFactory;
+	private NoSuchAlgorithmException trustManagerFactoryException;
+
 	public UrlConnectionHttpClient(String userAgent, String acceptedMediaType, String version){
 		this(userAgent, acceptedMediaType, version, null, null, null);
 	}
-
+	
 	public UrlConnectionHttpClient(String userAgent, String acceptedMediaType,
 			String version, ISSLCertificateCallback callback, Integer configTimeout, String excludedSSLCipherRegex) {
+		this(userAgent, acceptedMediaType, version, callback, configTimeout, excludedSSLCipherRegex, null, null);
+	}
+
+	public UrlConnectionHttpClient(String userAgent, String acceptedMediaType,
+			String version, ISSLCertificateCallback callback, Integer configTimeout, String excludedSSLCipherRegex,
+			String alias, X509Certificate cert) {
 		this.userAgent = userAgent;
 		this.acceptedMediaType = acceptedMediaType;
 		this.acceptedVersion = version;
 		this.sslAuthorizationCallback = callback;
 		this.configTimeout = configTimeout;
 		this.excludedSSLCipherRegex = excludedSSLCipherRegex;
+		this.initTrustManagerFactory(alias, cert);
+	}
+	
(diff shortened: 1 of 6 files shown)


src/main/java/com/openshift/internal/restclient/http/UrlConnectionHttpClient.java
```diff
@@ -69,18 +69,60 @@
...
+	public UrlConnectionHttpClient(String userAgent, String acceptedMediaType,
+			String version, ISSLCertificateCallback callback, Integer configTimeout, String excludedSSLCipherRegex,
+			String alias, X509Certificate cert) {
 		this.userAgent = userAgent;
 		this.acceptedMediaType = acceptedMediaType;
 		this.acceptedVersion = version;
 		this.sslAuthorizationCallback = callback;
 		this.configTimeout = configTimeout;
 		this.excludedSSLCipherRegex = excludedSSLCipherRegex;
+		try {
```
jcantrill: I would refactor this to a separate method like 'initTrustManagerFactory'

src/main/java/com/openshift/internal/restclient/http/UrlConnectionHttpClientBuilder.java
```diff
@@ -44,6 +48,12 @@ public UrlConnectionHttpClientBuilder setCredentials(String username, String pas
 		return setAuthorizationStrategy(new BasicAuthorizationStrategy(username, password, token));
 	}
 	
+	public UrlConnectionHttpClientBuilder setCertificate(String alias, X509Certificate cert) {
```
jcantrill: My knowledge of certs is limited so I'm just asking, would we always expect this to be pairs of alias/cert?  opposed to providing 2 separate builder methods.  Also, is there any thought if there is need or if it makes sense for a consumer to call this multiple times for more then one cert?

jcantrill:
other then my comments LGTM.  @adietish  you have any comments?
[test]

src/main/java/com/openshift/internal/restclient/http/UrlConnectionHttpClient.java
gabemontero: good point - will do

src/main/java/com/openshift/internal/restclient/http/UrlConnectionHttpClientBuilder.java
gabemontero: Definitely based on the KeyStore API it is a alias/cert pair.

My experience isn't significant either, but so far the client side scenario has been a single cert.  But I'm not necessarily opposed to say a Map<String,X509Certificate> input parameter structure if you feel strongly enough about it.

gabemontero:
update to @jcantrill first comment pushed

src/main/java/com/openshift/internal/restclient/http/UrlConnectionHttpClientBuilder.java
jcantrill: I dont have strong oppinions

jcantrill:
[merge]

PR #127, opened 2016-03-10
[JBIDE-21559] Retrieve specific container logs. Fix dc#setReplicaSele…

jcantrill:
…ctor

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/capability/resources/OpenShiftBinaryPodLogRetrieval.java
+++ b/src/main/java/com/openshift/internal/restclient/capability/resources/OpenShiftBinaryPodLogRetrieval.java
@@ -12,22 +12,30 @@ package com.openshift.internal.restclient.capability.resources;
 
 import java.io.InputStream;
 import java.io.SequenceInputStream;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Consumer;
 
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.openshift.restclient.IClient;
 import com.openshift.restclient.capability.resources.IPodLogRetrieval;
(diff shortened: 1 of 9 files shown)


src/main/java/com/openshift/internal/restclient/capability/resources/OpenShiftBinaryPodLogRetrieval.java
```diff
@@ -39,10 +41,16 @@ public boolean isSupported() {
...
 	}
-
+	
 	@Override
 	public InputStream getLogs(boolean follow) {
+		return getLogs(follow, null);
+	}
+
+	@Override
+	public InputStream getLogs(boolean follow, String container) {
```
scabanovich: Can this method be called concurrently, the second call changing the inner state while the first call has not yet requested build args?

jcantrill: Yes, there is that possibility, but it was there before as well.  I'll investigate how to address.

jcantrill:
[test]

jcantrill:
@scabanovich updated PR

src/main/java/com/openshift/internal/restclient/capability/resources/OpenShiftBinaryPodLogRetrieval.java
```diff
@@ -39,42 +47,114 @@ public boolean isSupported() {
...
+		}
+
+		@Override
+		protected void cleanup() {
+			follow = false;
+			if(getProcess() != null) {
+				IOUtils.closeQuietly(getProcess().getInputStream());
+				IOUtils.closeQuietly(getProcess().getErrorStream());
+			}
+			cacheFlush.accept(this.container);
```
scabanovich: It seems that this is not a synchronized call to cache.remove. Class PodLogs is inner in OpenShiftBinaryPodLogRetrieval so that it sees cache and can call 
synchronized(cache) { cache.remove(container); }

jcantrill:
[merge]

PR #135, opened 2016-03-23
JBIDE-21453 - Server Adapter: use rsync flag instead of local filtered-copy

xcoulon:
Add the "--exclude='.git'" flag in the rsync command to exclude the '.git'
 directory. Since there can be only one occurrence of this option,
 excluding the '.git' directory seems like the best choice.

Also refactored the methods that generate the command line flags to simply return
 a String that will be appended to the StringBuilder by the caller method.

Signed-off-by: Xavier Coulon <xcoulon@redhat.com>


adietish:
+1, looks good to me

adietish:
[merge]

PR #137, opened 2016-03-24
JBIDE-21453 - Server Adapter: use rsync flag instead of local filtered-copy

xcoulon:
Cleaning code: removing outdated comments and unused methods and imports.


adietish:
[merge]

PR #138, opened 2016-03-29
JBIDE-21981 - Error during rsync

xcoulon:
The OpenShiftBinaryRSync#sync(Peer,Peer) returned an InputStream
that was the combination of the underlying Process#getInputStream()
with Process#getErrorStream(), which means that when an error occurred,
the process' error stream was already read and consumed, meaning that it
could not be read again when throwing an OpenShiftException.

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/capability/resources/AbstractOpenShiftBinaryCapability.java
+++ b/src/main/java/com/openshift/internal/restclient/capability/resources/AbstractOpenShiftBinaryCapability.java
@@ -11,6 +11,8 @@
 package com.openshift.internal.restclient.capability.resources;
 
 import java.io.IOException;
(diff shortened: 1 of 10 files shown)


fbricon:
It'd be nice(r) we had a way to pass rsync flags via API.

xcoulon:
ok, I'll add some varargs on the sync(Peer, Peer) method to specify some options, then.

xcoulon:
@fbricon @adietish externalized some options to the `oc' command that will be run by the client. Please look at https://github.com/jbosstools/jbosstools-openshift/pull/1064 for the impact on the Eclipse tooling.

xcoulon:
Closing this issue and reopening with a different JIRA issue number.

PR #139, opened 2016-03-29
Catch http.Unauthorized exception as well

jcantrill:


jcantrill:
[test]  cc @adietish while debugging I saw an issue where we were throwing the previous version of this exception but not catching it.

adietish:
nice catch!
do you know why we're throwing the legacy exception?

jcantrill:
@adietish  I do not know ATM.  I thought I had caught all instances in the refactoring.
[merge]

PR #140, opened 2016-03-30
JBIDE-21999 - Error during rsync

xcoulon:
The OpenShiftBinaryRSync#sync(Peer,Peer) returned an InputStream
 that was the combination of the underlying Process#getInputStream()
 with Process#getErrorStream(), which means that when an error occurred,
 the process' error stream was already read and consumed, meaning that it
 could not be read again when throwing an OpenShiftException.

Also, added an 'OpenShiftBinaryOption' enum to hold the optional arguments
 to pass to the underlying 'oc' command, which means that it impacts all
 3 classes implementing calls to 'oc' (rsync, port-forwarding and logs)
The side effect to this commit is that by default, no flags are passed.

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/capability/resources/AbstractOpenShiftBinaryCapability.java
+++ b/src/main/java/com/openshift/internal/restclient/capability/resources/AbstractOpenShiftBinaryCapability.java
@@ -11,6 +11,8 @@
 package com.openshift.internal.restclient.capability.resources;
 
 import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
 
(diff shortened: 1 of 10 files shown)


adietish:
as discussed we'll keep the enhancement that allows users to accept/refute the certificate to later.

adietish:
enum seem like the right tool if the number of options is strictly limited. OpenShiftBinaryOption seem not limited so I'd think that an enum is not the right tool at hand. But we can keep this for later given that we have strict time constraints

adietish:
other than that +1

adietish:
[merge]

xcoulon:
thanks, @adietish !

PR #145, opened 2016-04-06
[JBIDE-22127] chanch WatchClient synch blocks to minimize blocking

jcantrill:
ptl @adietish @fbricon

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/WatchClient.java
+++ b/src/main/java/com/openshift/internal/restclient/WatchClient.java
@@ -16,8 +16,14 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.commons.lang.StringUtils;
+import org.eclipse.jetty.util.component.LifeCycle;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.eclipse.jetty.websocket.api.Session;
 import org.eclipse.jetty.websocket.api.UpgradeException;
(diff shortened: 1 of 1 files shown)


jcantrill:
[test]

fbricon:
Seems much faster now and I couldn't reproduce any blocking issue. So that counts as +1 from me.

@scabanovich, @alexeykazakov, @adietish, @jeffmaury  any 2nd opinion on this change?

adietish:
this requires us to deploy a 4.0.2.Final artifact to repository.jboss.org and reference it in JBoss Tools.

src/main/java/com/openshift/internal/restclient/WatchClient.java
```diff
@@ -52,9 +58,68 @@
...
+		Stopping
+	}
 	
 	static {
-		wsClient = newWebSocketClient();
+		try {
+			lockTimeout = Long.parseLong(System.getProperty("com.openshift.restclient.watchlocktimeoutms", "30000")); 
+		}catch(NumberFormatException e) {
+			lockTimeout = 30 * 1000;
+		}
```
adietish: can we please extract this try/catch-block to a method with a speaking name like #getLockTimeout?

src/main/java/com/openshift/internal/restclient/WatchClient.java
```diff
@@ -52,9 +58,68 @@
...
+				public void lifeCycleFailure(LifeCycle event, Throwable cause) {
+					LOGGER.error("The watchclient failed:", cause);
+					status.set(Status.Stopped);
+				}
+			});
+			
+		}catch(Exception e) {
+			LOGGER.error("Exception initializing watch client",e);
+			throw e;
+		}
```
adietish: can we extract this try/catch-block to a method with a speaking name like #createWebSocketClient?

adietish: Logging + throwing... wont this end in logging twice since the one user that's launching this will most likely also have a logging statement in his catch block?

adietish: generally speaking it's not very nice to have code in a static initializer which can throw Exceptions. But the possible alternative, to have it created in a lazy manner requires locking which we dont want to add in such a late change.

adietish:
i made my suggested changes in #146

src/main/java/com/openshift/internal/restclient/WatchClient.java
```diff
@@ -136,30 +201,31 @@ public IWatcher watch(Collection<String> kinds, String namespace, IOpenShiftWatc
...
 	
 	public void start() {
-		synchronized (wsClient) {
-			if(wsClient.isStarted() || wsClient.isStarting()) return;
-			try {
-				wsClient.start();
-			} catch (Exception e) {
-				throw createOpenShiftException(String.format("Could not start watchClient"),e);
-			}
+		if(status.get().equals(Status.Started) || status.get().equals(Status.Starting)) {
```
adietish: enums should be compared with ==, no #equals, like you already do when checking the status for Status.STARTING (line 210)

adietish:
supplanted by #146 with minor changes

PR #148, opened 2016-04-18
fix sanitize to allow keys with empty content to handle port specs

jcantrill:
[test] @fbricon @adietish 

during working on build from found an issue in our method to sanitize the json that was keeping us from getting port spec info from imagestream tags since all the relevant info is in the key.


jcantrill:
[merge]

PR #150, opened 2016-04-19
allow BuildConfig creation via fluent style builder

jcantrill:
@fbricon @adietish 
This PR allows for a fluent style building of a buildconfig used in the create from source workflow in JBT.  Please review

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/model/BuildConfig.java
+++ b/src/main/java/com/openshift/internal/restclient/model/BuildConfig.java
@@ -50,6 +50,7 @@ public class BuildConfig extends KubernetesResource implements IBuildConfig {
 	private static final String BUILDCONFIG_SOURCE_TYPE = "spec.source.type";
 	private static final String BUILDCONFIG_SOURCE_URI = "spec.source.git.uri";
 	private static final String BUILDCONFIG_SOURCE_REF = "spec.source.git.ref";
+	
 	public static final String BUILDCONFIG_TYPE = "spec.strategy.type";
 	private static final String BUILDCONFIG_CUSTOM_IMAGE = "spec.strategy.customStrategy.image";
 	private static final String BUILDCONFIG_CUSTOM_EXPOSEDOCKERSOCKET = "spec.strategy.customStrategy.exposeDockerSocket";
@@ -75,8 +76,7 @@ public class BuildConfig extends KubernetesResource implements IBuildConfig {
 		super(node, client, null);
 		CapabilityInitializer.initializeCapabilities(getModifiableCapabilities(), this, client);
 	}
-	
-	
+
 	@Override
 	public IObjectReference getBuildOutputReference() {
 		ModelNode node = get("spec.output.to");
@@ -91,7 +91,7 @@ public class BuildConfig extends KubernetesResource implements IBuildConfig {
 		List<ModelNode> list = get(BUILDCONFIG_TRIGGERS).asList();
 		final String url = getClient() != null ? getClient().getResourceURI(this) : "";
 		for (ModelNode node : list) {
-			String type = node.get("type").asString();
+			String type = node.get(TYPE).asString();
 			switch(type){
 				case BuildTriggerType.generic:
 				case BuildTriggerType.GENERIC:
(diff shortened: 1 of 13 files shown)


src/main/java/com/openshift/internal/restclient/model/build/BuildConfigBuilder.java
```diff
@@ -0,0 +1,176 @@
...
+		private IBuildConfigBuilder builder;
+		private GitBuildSource source;
+		
+		GitSourceBuilder(IBuildConfigBuilder builder, GitBuildSource source){
+			this.builder = builder;
+			this.source = source;
+		}
+
+		@Override
+		public IBuildConfigBuilder end() {
```
adietish: better maybe call it #build for API consistency?

adietish: maybe better call it #build for API consistency?

adietish:
I would warmly recommend to add tests for this since those help in code maintenance and as a poor man documentation

src/main/java/com/openshift/internal/restclient/model/build/BuildConfigBuilder.java
adietish: calling this method #end seems caused by the fact that this isnt a builder really. It applies the settings directly whereas a builder would store the settings locally and apply them once you call #build. Not all happy with this unexpected naming but I dont have any better suggestion right now.

jcantrill: This follows a pattern of what you are able to do with the f8 client.  This pattern is much nicer to work with then the setXXXXX that is required otherwise.  They impl 'end' to imply  you can continue to build up the instance without the final construction.  I would be inclined to change the impl to store local values before actually 'building' the various instance if that would be more agreeable to you.

jcantrill:
[test] 
Add test for @adietish

src/main/java/com/openshift/internal/restclient/model/BuildConfig.java
```diff
@@ -138,17 +140,22 @@ public void addBuildTrigger(IBuildTrigger trigger) {
...
+		case BuildTriggerType.IMAGE_CHANGE:{
 			if(!(trigger instanceof IImageChangeTrigger)) {
 				throw new IllegalArgumentException("IBuildTrigger of type imageChange does not implement IImageChangeTrigger");
 			}
 			IImageChangeTrigger image = (IImageChangeTrigger)trigger;
-			triggerNode.get(getPath(BUILD_CONFIG_IMAGECHANGE_IMAGE)).set(defaultIfNull(image.getImage()));
-			triggerNode.get(getPath(BUILD_CONFIG_IMAGECHANGE_NAME)).set(defaultIfNull(image.getFrom()));
-			triggerNode.get(getPath(BUILD_CONFIG_IMAGECHANGE_TAG)).set(StringUtils.defaultIfBlank(image.getTag(), ""));
+			if(image.getImage() != null)
+				triggerNode.get(getPath(BUILD_CONFIG_IMAGECHANGE_IMAGE)).set(defaultIfNull(image.getImage()));
```
fbricon: just nitpicking here: don't need defaulIfNull since null check is already done :-)

Same goes for the next lines of code

src/main/java/com/openshift/internal/restclient/model/build/SourceBuildStrategy.java
```diff
@@ -87,6 +110,7 @@ public void setScriptsLocation(String location) {
 
 	@Override
 	public void setEnvVars(Collection<IEnvironmentVariable> envVars) {
+		if(envVars == null) return;
```
fbricon: or should we clear the existing env vars?

fbricon: Either way, that should be javadoc'ed

fbricon:
Except for really minor comment, that's looks ok to me

jcantrill:
updated per @fbricon  nits
[merge]

PR #158, opened 2016-05-19
Add ConfigMap resource type

lulf:


lulf:
This might be a bit premature. I suppose you might want to have the appropriate types for the ConfigMap resource itself too.

PR #164, opened 2016-05-24
[OSJC-256] DefaultClient#equal <true> if authstrategies match in username

adietish:


adietish:
[test]

adietish:
@jcantrill dare to review?

adietish:
[test]

adietish:
weird, travis fails to build complainig that it cannot find the branch OSJC-256...

src/test/java/com/openshift/internal/restclient/DefaultClientTest.java
```diff
@@ -165,18 +165,6 @@ public void clientShouldNotEqualClientWithDifferentUrl() throws Exception {
...
-		tokenClient.setAuthorizationStrategy(new TokenAuthorizationStrategy("aToken", "aUser"));
-
-		DefaultClient basicAuthClient = new DefaultClient(baseUrl, null);
-		basicAuthClient.setAuthorizationStrategy(new BasicAuthorizationStrategy("aUser", "aPassword", "aPassword"));
-
-		assertThat(tokenClient).isNotEqualTo(basicAuthClient);
-	}
-
-	@Test
 	public void client_should_equal_client_with_same_TokenAuthStrategy_with_different_token() throws Exception {
```
jcantrill: @adietish Looks like you lost a test

jcantrill:
LGTM.  Per our conversation, I think equality should be obtained when the username and server match.   Other then verifying the 'lost' test I think this is ok to merge.

src/test/java/com/openshift/internal/restclient/DefaultClientTest.java
adietish: @jcantrill nope, that's on purpose. See the method name, it says that 2 clients with different strategies should not be equal, that's exactly the opposite of what I wanted to change, so the test has to disappear.

adietish:
[merge]

PR #165, opened 2016-05-24
[OSJC-237] Add support for API groups and remove hardcoding of api ty…

jcantrill:
…pe mapping

cc @fbricon @adietish

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
+++ b/src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
@@ -1,0 +1,394 @@
+/******************************************************************************* 
+ * Copyright (c) 2016 Red Hat, Inc. 
+ * Distributed under license by Red Hat, Inc. All rights reserved. 
+ * This program is made available under the terms of the 
+ * Eclipse Public License v1.0 which accompanies this distribution, 
+ * and is available at http://www.eclipse.org/legal/epl-v10.html 
+ * 
+ * Contributors: 
+ * Red Hat, Inc. - initial API and implementation 
+ ******************************************************************************/
+package com.openshift.internal.restclient;
+
+import java.net.MalformedURLException;
+import java.net.SocketTimeoutException;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.apache.commons.lang.StringUtils;
+import org.jboss.dmr.ModelNode;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.openshift.internal.restclient.http.HttpClientException;
+import com.openshift.internal.restclient.http.NotFoundException;
+import com.openshift.internal.restclient.model.properties.ResourcePropertyKeys;
+import com.openshift.internal.util.JBossDmrExtentions;
+import com.openshift.restclient.IApiTypeMapper;
+import com.openshift.restclient.OpenShiftException;
+import com.openshift.restclient.ResourceKind;
+import com.openshift.restclient.http.IHttpClient;
+import com.openshift.restclient.http.IHttpConstants;
+import com.openshift.restclient.model.IResource;
+
+/**
+ * Typemapper to determine the endpoints for
+ * various openshift resources
+ * @author jeff.cantrill
+ *
+ */
+public class ApiTypeMapper implements IApiTypeMapper, ResourcePropertyKeys{
+	
+	private static final String FWD_SLASH = "/";
+	private static final String KUBE_API = "api";
+	private static final String OS_API = "oapi";
+	private static final String API_GROUPS_API = "apis";
+	private static final Logger LOGGER = LoggerFactory.getLogger(ApiTypeMapper.class);
+	private final String baseUrl;
+	private final IHttpClient client;
+	private List<VersionedApiResource> resourceEndpoints;
+	private Map<String, String> preferedVersion = new HashMap<>(2);
+	
+	public ApiTypeMapper(String baseUrl, IHttpClient client) {
+		this.baseUrl = baseUrl;
+		this.client = client;
+		preferedVersion.put(KUBE_API, KubernetesAPIVersion.v1.toString());
+		preferedVersion.put(OS_API, OpenShiftAPIVersion.v1.toString());
+	}
+
+	@Override
+	public boolean isSupported(IResource resource) {
+		return isSupported(resource.getApiVersion(), resource.getKind());
+	}
+	
+	@Override
+	public IVersionedApiResource getEndpointFor(String apiVersion, String kind) {
+		init();
+		IVersionedApiResource apiresource = endpointFor(apiVersion, kind);
+		if(apiresource == null) {
+			throw new OpenShiftException("No endpoint found for %s, version %s", kind, apiVersion);
+		}
+		return apiresource;
+	}
+	
+	private IVersionedApiResource endpointFor(String version, String kind) {
+		String[] split = StringUtils.isBlank(version) ? new String [] {} : version.split(FWD_SLASH);
+		Optional<IVersionedApiResource> result = null;
+		if(split.length <=1) {
+			result = Stream.of(KUBE_API, OS_API)
+					.map(api->formatEndpointFor(api, (split.length == 0 ? preferedVersion.get(api) : split[0]), kind))
+					.filter(e->resourceEndpoints.contains(e))
+					.findFirst();
+		}else{
+			result = Optional.of(formatEndpointFor(API_GROUPS_API, version, ResourceKind.pluralize(kind, true, true)));
+		}
+		if(result.isPresent()) {
+			int index = resourceEndpoints.indexOf(result.get());
+			if(index > -1) {
+				return resourceEndpoints.get(index);
+			}
+		}
+		return null;
+	}
+
+	@Override
+	public boolean isSupported(String kind) {
+		return isSupported(null, kind);
+	}
+
+	@Override
+	public boolean isSupported(String version, String kind) {
+		init();
+		return endpointFor(version, kind) != null;
+	}
+	
+	private IVersionedApiResource formatEndpointFor(String prefix, String version, String kind) {
+		return new VersionedApiResource(prefix, version, ResourceKind.pluralize(kind, true, true));
+	}
+	
+	private synchronized void init() {
+		if(resourceEndpoints != null) {
+			return;
+		}
+		resourceEndpoints = new ArrayList<>();
+		Collection<ApiGroup> groups = getLegacyGroups();
+		groups.addAll(getApiGroups());
+		groups.forEach(g->{
+			Collection<String> versions = g.getVersions();
+			versions.forEach(v->{
+				Collection<ModelNode> resources = getResources(g, v);
+				addEndpoints(g.getPrefix(), g.getName(), v, resources);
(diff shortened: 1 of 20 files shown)


jcantrill:
[test]

jcantrill:
[test]

src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
```diff
@@ -0,0 +1,389 @@
...
+		private final String prefix;
+		private final String name;
+		private final boolean namespaced;
+		private final Collection<String> capabilities = new ArrayList<>();
+		private final String version;
+		private String apiGroupName;
+		private String kind;
+		
+		VersionedApiResource(String prefix, String version, String name){
+			if(version.contains(FWD_SLASH)) {
```
adietish: what if version is null?

src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
```diff
@@ -0,0 +1,389 @@
...
+				return resourceEndpoints.get(index);
+			}
+		}
+		return null;
+	}
+
+
+	@Override
+	public boolean isSupported(String version, String kind) {
+		init();
```
adietish: apparently #isSupported is required to be called in the 1st place (since it calls #init). Shouldnt we therefore assert in the other methods that #init was called?

src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
```diff
@@ -0,0 +1,389 @@
...
+	@Override
+	public boolean isSupported(IResource resource) {
+		return isSupported(resource.getApiVersion(), resource.getKind());
+	}
+	
+	@Override
+	public IVersionedApiResource getEndpointFor(String apiVersion, String kind) {
+		IVersionedApiResource apiresource = endpointFor(apiVersion, kind);
+		if(apiresource == null) {
+			throw new OpenShiftException("No endpoint found for kind %s", kind);
```
adietish: Shouldnt the error also state the version that the kind was not supported in?

src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	@Override
+	public boolean isSupported(IResource resource) {
+		return isSupported(resource.getApiVersion(), resource.getKind());
+	}
+	
+	@Override
+	public IVersionedApiResource getEndpointFor(String apiVersion, String kind) {
+		IVersionedApiResource apiresource = endpointFor(apiVersion, kind);
```
adietish: Shouldnt we assert that #init was called?

src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
```diff
@@ -0,0 +1,389 @@
...
+	
+	private String readEndpoint(final String endpoint) {
+		try {
+			final URL url = new URL(new URL(this.baseUrl), endpoint);
+			LOGGER.debug(url.toString());
+			String response = client.get(url, IHttpConstants.DEFAULT_READ_TIMEOUT);
+			LOGGER.debug(response);
+			return response;
+		} catch (MalformedURLException | SocketTimeoutException e) {
+			LOGGER.error("Exception", e);
```
adietish: Logging and throw is imho an anti-pattern since the caller will most likely log an exception that he catches.

src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
```diff
@@ -0,0 +1,389 @@
...
+			throw new OpenShiftException(e,"");
+		//HACK - This gets us around a server issue
+		} catch (HttpClientException e) {
+			if(e instanceof NotFoundException) {
+				throw new com.openshift.restclient.NotFoundException(e);
+			}
+			LOGGER.error("Unauthorized exception. Can system:anonymous get the API endpoint", e);
```
adietish: see above

src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
```diff
@@ -0,0 +1,389 @@
...
+	static class ApiGroup implements IApiGroup{
+		private final ModelNode node;
+		private final String prefix;
+		private final String path;
+		
+		ApiGroup(String prefix, ModelNode node) {
+			this.prefix = prefix;
+			this.node = node;
+			StringBuilder builder = new StringBuilder(prefix);
+			if(getName() != null) { //null name for k8e or openshift
```
adietish: calling #getName() is dangerous since you implicitly require the node to be initialized beforehand (something that you only discover when reading the implementation of the method). I'd most likely provide a protected method with "node" as parameter, calling it with the instance variable for the public parameterless variant. Something like:
```
protected String getName(ModelNode node) {
	return JBossDmrExtentions.asString(node, new HashMap<>(), "name");
}
public String getName() {
	return getName(node);
}
```
and then calling
```
getName(node)
```
here.

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -74,22 +75,24 @@
...
 	
-	private final Map<String, String> typeMappings = new HashMap<String, String>();
 	private String openShiftVersion;
 	private String kubernetesVersion;
 	private IAuthorizationStrategy strategy;
 	private IAuthorizationClient authClient;
+	private IApiTypeMapper typeMapper;
 	
 	@Deprecated
 	public DefaultClient(URL baseUrl, ISSLCertificateCallback sslCertCallback){
```
adietish: I would take the occasion to clean this deprecated method (remove it?)

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -143,10 +147,10 @@ public String getResourceURI(IResource resource) {
 	@Override
 	public <T extends IResource> List<T> list(String kind, String namespace, Map<String, String> labels) {
 		try {
-			if(!getTypeMappings().containsKey(kind))
+			if(!typeMapper.isSupported(null, kind))
```
adietish: apparently passing "null" as version has a special meaning. I would document this in javadoc

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -204,7 +208,7 @@ public String getResourceURI(IResource resource) {
 		if(ResourceKind.LIST.equals(resource.getKind())) throw new UnsupportedOperationException("Generic create operation not supported for resource type 'List'");
 		try {
 			namespace = ResourceKind.PROJECT.equals(resource.getKind()) ? "" : namespace;
```
adietish: Shouldnt we use "null" instead of "" as a namespace if we dont use it? I see that URLBuilder will accept both (empty or null - it's using StringUtils#isBlank). But in my understanding null is more obvious than an empty string for stating that the namespace shall not be used.

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -223,7 +227,7 @@ public String getResourceURI(IResource resource) {
 		if(ResourceKind.LIST.equals(kind)) throw new UnsupportedOperationException("Generic create operation not supported for resource type 'List'");
 		try {
 			namespace = ResourceKind.PROJECT.equals(kind) ? "" : namespace;
```
adietish: I see the same logic (use "" for namespace if we're about to build the url for a project) being repeated several times. Shouldnt we thus extract it to a single, reusable method (so that it gets easier to correct in a single location)?

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -262,7 +266,7 @@ public String getResourceURI(IResource resource) {
 		if(ResourceKind.LIST.equals(resource.getKind())) throw new UnsupportedOperationException("Delete operation not supported for resource type 'List'");
 		try {
 			String namespace = ResourceKind.PROJECT.equals(resource.getKind()) ? "" : resource.getNamespace();
```
adietish: see above

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -300,7 +304,7 @@ public IList get(String kind, String namespace) {
 	public <T extends IResource> T get(String kind, String name, String namespace) {
 		try {
 			namespace = ResourceKind.PROJECT.equals(kind) ? "" : namespace;
```
adietish: see above

adietish:
quite a large PR, hard to get the overall architecture but I commented what I found in a first tour of it.

src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
jcantrill: fixed

src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
jcantrill: added init to the other public method.

src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
jcantrill: fixed

src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
jcantrill: removed log message

src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
jcantrill: here i will defer not replacing unless we want to wrap up the exception in openshiftexception with the message that we are logging here.

src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
jcantrill: What about moving this logic to an initialization method?  or an init for the path or 'setPath(node)'.  I kind of find needing to look at the impl of 'getName()' irrelevant since really all we are saying is to retrieve the name and that is how we will create the path item

src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
jcantrill: it can't be null otherwise there is an error in the response from the server.  We could add a throw or assert which would check for null.  You think this would be a better option?

src/main/java/com/openshift/internal/restclient/DefaultClient.java
jcantrill: removed

src/main/java/com/openshift/internal/restclient/DefaultClient.java
jcantrill: it is documented.  should we just overload the method?  would that be better?

src/main/java/com/openshift/internal/restclient/DefaultClient.java
jcantrill: I think this is code left from the early impl of the class when the server api was flaky. Using null might make more sense here but I have not thought about it in depth

src/main/java/com/openshift/internal/restclient/DefaultClient.java
jcantrill: overloaded the method

jcantrill:
[test]

jcantrill:
bumped the pom to 4.1.0 per request from @fbricon   [merge]

PR #167, opened 2016-05-26
[JBIDE-22423] fix unable to edit buildconfigs by only retrieving from…

jcantrill:
… nodes when the  key exists

cc @fbricon   This fixes the issue by only retrieving keys when they already exist.  DMR allows you to get().set() which results in the key being added to the tree.  There are cases where this is undesirable.


fbricon:
Do you have any unit tests proving it works? ;-)

fbricon:
other than the missing test, the fix seems to work as advertised. All ITs in jbosstools-openshift still pass.

jcantrill:
@fbricon  added test.  [merge]

PR #168, opened 2016-05-31
Allow empty lists in responses

lulf:
Even though empty lists are required for some parts of the API, other parts allow lists to not be
defined. This patch changes the behavior when the list is not defined by returning an empty list
rather than failing.

It would be nice to still get failure when the list is required, but that would require some more rework of the client API. Let me know what you think.


jcantrill:
LGTM [merge]

PR #169, opened 2016-06-01
JBIDE-22473 Cannot remove env vars

scabanovich:


src/main/java/com/openshift/internal/restclient/model/ReplicationController.java
```diff
@@ -99,6 +99,44 @@ public void setEnvironmentVariable(String containerName, String name, String val
...
+				ModelNode node =  opt.isPresent() ? opt.get() : containers.get(0);
+				ModelNode envNode = get(node, ENV);
+
+				List<ModelNode> varList = new ArrayList<>();
+				if (ModelType.LIST.equals(envNode.getType())){
+					varList.addAll(envNode.asList());
+				}
+
+				//Check if variable exists
+				Optional<ModelNode> targetVar = varList.stream().filter(n->name.equals(asString(n, NAME))).findFirst();
```
jcantrill: probably should do a checko for null name here

jcantrill:
[test]

src/main/java/com/openshift/internal/restclient/model/ReplicationController.java
scabanovich: Since null name will not do any good, can it be checked in the first line of the method? What is better, return silently or throw IllegalArgumentException?

jcantrill: Probably better to throw some sort of runtime exception then swallow silently

scabanovich: So the method itself should not throw an exception? Then I suggest just
if(name == null) {
  return;
}
in the beginning of the method, because no var to be removed will be selected anyway. Is it ok?

jcantrill: I was suggesting we throw the exception and not return silently

scabanovich: Thank you, done. I misspell 'than' and 'then' 90% of times, bad for me to read that word in your previous comment literally. :)

adietish:
@jcantrill you commented in jira that the removal wont have any effect if the var is defined in an image. Is this check required or something we can postpone for later?

adietish:
[merge]

PR #175, opened 2016-06-20
fix case where error is thrown while trying to initialize ApiTypeMapper

jcantrill:
fix issue where an error is thrown during initialization which caused subsequent checks to not recognize valid types.

cc @fbricon @adietish 

[test]

### DIFF
--- a/src/test/java/com/openshift/internal/restclient/ApiTypeMapperTest.java
+++ b/src/test/java/com/openshift/internal/restclient/ApiTypeMapperTest.java
@@ -11,7 +11,11 @@
 package com.openshift.internal.restclient;
 
 import static org.junit.Assert.*;
+import static org.mockito.Matchers.anyInt;
+import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.*;
+
+import java.net.URL;
 
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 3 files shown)


src/test/java/com/openshift/internal/restclient/ApiTypeMapperTest.java
```diff
@@ -32,10 +36,22 @@
 	private MocksFactory factory = new MocksFactory();
 	
 	@Test
+	public void testKubernetesResourceIsSupportedAfterInitiallyErrorIsThrown() throws Exception {
+		IService resource = factory.stub(ResourceKind.SERVICE);
+		try {
+			when(getHttpClient().get(eq(new URL(super.base + "/api")), anyInt())).thenThrow(new RuntimeException());
+			assertTrue("Exp. Kube support", mapper.isSupported(resource));
+		}catch(RuntimeException e) {
```
fbricon: the test can't cover both paths. It's either one or the other

jcantrill: @fbricon I didnt catch this before the merge. What do you mean it can't cover 'both' paths.  What is 'both' in this sentence.  The test is meant to catch the case where the internal collection is initialized before it is populated and then subsequent actions are unable to find the supported endpoints.

PR #178, opened 2016-06-21
Allow secret volumes to be added to replication controllers. 

grs:
This is only exposed through internal object. I realise there is some work in progress on volumes which will influence the public API when complete.

### DIFF
--- a/src/test/java/com/openshift/internal/restclient/model/v1/ReplicationControllerTest.java
+++ b/src/test/java/com/openshift/internal/restclient/model/v1/ReplicationControllerTest.java
@@ -294,6 +294,22 @@ public class ReplicationControllerTest {
 		
 		JSONAssert.assertEquals(exp.toJSONString(false), container.toJSONString(), true);
 	}
-	
 
+	@Test
+	public void testAddSecretVolumeToPodSpec() throws JSONException {
+            IVolumeMount volumeMount = new IVolumeMount() {
+                    public String getName() { return "my-secret"; }
+                    public String getMountPath() { return "/path/to/my/secret/"; }
+                    public boolean isReadOnly() { return true; }
+                    public void setName(String name) {}
+                    public void setMountPath(String path) {}
+                    public void setReadOnly(boolean readonly) {}
+                };
+            ((ReplicationController) rc).addSecretVolumeToPodSpec(volumeMount, "the-secret");
+            Set<IVolumeSource> podSpecVolumes = rc.getVolumes();
+            Optional vol = podSpecVolumes.stream()
+                .filter(v->v.getName().equals(volumeMount.getName()))
+                .findFirst();
+            assertTrue("Expected to find secret volume in pod spec", vol.isPresent());
+        }
 }
--- a/src/main/java/com/openshift/internal/restclient/model/ReplicationController.java
+++ b/src/main/java/com/openshift/internal/restclient/model/ReplicationController.java
@@ -283,22 +283,40 @@ public class ReplicationController extends KubernetesResource implements IReplic
 		return container;
 	}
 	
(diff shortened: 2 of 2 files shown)


src/test/java/com/openshift/internal/restclient/model/v1/ReplicationControllerTest.java
```diff
@@ -294,6 +294,26 @@ public void testAddContainerAllowsContainerToBeFurtherManipulated()  throws JSON
...
+                    public String getName() { return "my-secret"; }
+                    public String getMountPath() { return "/path/to/my/secret/"; }
+                    public boolean isReadOnly() { return true; }
+                    public void setName(String name) {}
+                    public void setMountPath(String path) {}
+                    public void setReadOnly(boolean readonly) {}
+                };
+            ((ReplicationController) rc).addSecretVolumeToPodSpec(volumeMount, "the-secret");
+            Set<IVolumeSource> podSpecVolumes = rc.getVolumes();
+            boolean found = false;
```
jcantrill: This might be cleaner with a stream, something like:
`
Optional<IVolumeSource> vol = podSpecVolumes.stream()
     .filter(v->v.getName().equals(volumeMount.getName()))
      .findFirst();

assertTrue("Expected to find secret volume in pod spec", vol.isPresent());
`

jcantrill:
Is the secret volume spec in such flux for v1 that we do not want to add a method to the interface?  The public interface methods are purposely designed in such a way that hopefully where things live in the JSON node tree are mostly irrelevant.  If we have the need to add secret volumes in some capacity now and in the future it would seem to make sense to me to make the interface public too.

cc @adietish 

[test]

src/test/java/com/openshift/internal/restclient/model/v1/ReplicationControllerTest.java
grs: I've updated the pr with this change. Thanks!

grs:
@jcantrill, sorry I realise my original comment was a bit ambiguous. What I meant was that it seemed that the API for this client had some work in progress around volumes, and that when that completes it may influence how secret volumes are exposed.

jcantrill:
Most if not all the exposure of the Api is based on needs of Jbosstools.
Feel free to suggest design changes and improvements. All contributions and
suggestions are welcome

On Tuesday, June 21, 2016, grs <notifications@github.com> wrote:

> @jcantrill <https://github.com/jcantrill>, sorry I realise my original
> comment was a bit ambiguous. What I meant was that it seemed that the API
> for this client had some work in progress around volumes, and that when
> that completes it may influence how secret volumes are exposed.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/openshift/openshift-restclient-java/pull/178#issuecomment-227529627>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AEVnOJxI-vu9N4T9c7J9SIQjcdWdTcVdks5qOC2egaJpZM4I69SD>
> .
>


-- 
--
Jeff Cantrill
OpenShift Integration Services
Red Hat, Inc.
jcantril@redhat.com
http://www.redhat.com

grs:
I could add the new method to the public interface if that works for you?

jcantrill:
@grs I have reviewed your PR in more depth and I believe I agree with you that I would not add this as part of the public api; more volume work needs to be done.  There probably needs to be more thought on how to add this via the public interfaces since really a secret is a type of volume source and IVolume should reflect that.  Looking through the https://github.com/openshift/origin code there is even some dependencies on how we use volume sources now that don't really properly describe volumes generically.

[merge]

grs:
@jcantrill Thanks! We will likely have a patch in the near future for persistent volume claims as well, which would be a little bit more involved and could start to get a public api you feel comfortable with.

PR #179, opened 2016-06-23
Impl OkHttpClient. Refactor authorization context

jcantrill:
This PR provides:

* Squashing multiple underlying httpclient dependencies to single library of OkHttp
* Removes authorization strategies to rely on OkHttp and challange handlers (similiar to oc login)
* Redefines what the authorization context api of IClient
* Fixes or add various integration tests.


jcantrill:
[test]
Corresponding JBT changes are forth coming

cc @fbricon @adietish @gabemontero

jcantrill:
@gabemontero  anyway you could test that it still works with the cert changes you provided to the previous client.  I dont have a test now to do that.

gabemontero:
@jcantrill yeah i believe i can get to it today

gabemontero:
@jcantrill i've started, but with the changes around AuthorizationStrategy, I need to make changes to the jenkins-plugin first to get it to compile ... started looking at the new IAuthorizationDetails stuff;

hopefully can still get to validating the SSL / trust store stuff today, but the outlook is just a bit longer since it is not a drop in and play sort of thing.

jcantrill:
I will work on updating the readme.

On Fri, Jun 24, 2016 at 11:40 AM, Gabe Montero <notifications@github.com>
wrote:

> @jcantrill <https://github.com/jcantrill> i've started, but with the
> changes around AuthorizationStrategy, I need to make changes to the
> jenkins-plugin first to get it to compile ... started looking at the new
> IAuthorizationDetails stuff;
>
> hopefully can still get to validating the SSL / trust store stuff today,
> but the outlook is just a bit longer since it is not a drop in and play
> sort of thing.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/openshift/openshift-restclient-java/pull/179#issuecomment-228379970>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AEVnOPRZqddudV4Erm2B-Krq8oNRU9Uxks5qO_plgaJpZM4I9K_I>
> .
>



-- 
--
Jeff Cantrill
OpenShift Integration Services
Red Hat, Inc.
jcantril@redhat.com
http://www.redhat.com

gabemontero:
The ssl cert callback / trust manager piece is working.  My callback is not getting called when the cert is set (meaning the trust manager down at the openshift-restclient-java level is happy), and is getting called when the cert is not set (like when we are in skip-tls mode).

One sidebar:  not sure if it is just the differences in build envs, or my misunderstanding of maven/pom.xml, but I had to explicitly add an dependency to okhttp in my pom.xml .... the classes were not pulled in by the dependency noted in the openshift-restclient-java pom.xml, which is what I was expecting.  Though perhaps once I pull openshift-restclient-java from the maven repo vs. my locally built copy, that will go away.

Note: given the relative disruption caused by these changes, I'm currently of the mind to *NOT* move https://github.com/openshift/jenkins-plugin up to 4.1.0-SNAPSHOT until after 3.3 ships.  But of course we can always revisit as things progress.  @bparees FYI

jcantrill:
@gabemontero The osgi changes that go along with this PR for jbosstools are here https://github.com/jbosstools/jbosstools-openshift/pull/1239/files#diff-7e585cd2377dbc5eb683f39b6a535615R60   This client is not part of the target platform so they get packaged explicitly into the OSGI plugin which i think would be similiar to your issue.  Also, have you tried mvn with the -U option?  Maybe that would pull it in too; my maven understanding is weak as well.

jcantrill:
@fbricon @adietish thoughts on proceeding with a merge?

jcantrill:
[test]

pom.xml
```diff
@@ -192,26 +202,6 @@
...
-		    <groupId>org.apache.httpcomponents</groupId>
-		    <artifactId>httpclient</artifactId>
-		    <version>4.3.6</version>
-		</dependency>
-		<dependency>
-		    <groupId>org.apache.httpcomponents</groupId>
-		    <artifactId>httpcore</artifactId>
-		    <version>4.3.3</version>
-		</dependency>
-		<dependency>
```
adietish: yay, -3 clients, replaced by a single one. great!

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -66,65 +57,50 @@
...
-		return  new UrlConnectionHttpClientBuilder()
-		.setAcceptMediaType("application/json")
-		.setCertificate(alias, cert)
-		.setSSLCertificateCallback(sslCertCallback)
-		.setConfigTimeout(configTimeoutMillies)
-		.client();
-	}
 	
 	@Override
+	public IClient clone() {
```
adietish: if we provide #clone, shouldnt we added the marker interface Cloneable?

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -180,7 +142,7 @@ public String getResourceURI(IResource resource) {
 		List<IResource> results = new ArrayList<IResource>(list.getItems().size());
 		for (IResource resource : list.getItems()) {
 			try{
-				results.add(createVersion(resource, namespace, resource.getApiVersion()));
+				results.add(create(resource, namespace));
 			}catch(OpenShiftException e){
 				if(e.getStatus() != null){
```
adietish: maybe OpenShiftException#hasStatus() would add explicitness to the API (e.getStatus() == null is easy to use but user dev needs to think at status possibly being null and check it, while #hasStatus() would unveil this explicitly right in the API)

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -200,122 +162,104 @@ public String getResourceURI(IResource resource) {
...
+		PUT,
+		POST,
+		DELETE
+	}
+	
+	@SuppressWarnings("unchecked")
+	private <T extends IResource> T execute(HttpMethod method, String kind, String namespace, String name, String subresource, IResource payload) {
+		if(ResourceKind.LIST.equals(kind)) 
+			throw new UnsupportedOperationException("Generic create operation not supported for resource type 'List'");
+		namespace = ResourceKind.PROJECT.equals(kind) ? null : namespace;
```
adietish: Shouldnt this be in URLBuilder?

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -200,122 +162,104 @@ public String getResourceURI(IResource resource) {
...
-			return factory.create(response);
-		} catch (HttpClientException e){
-			throw createOpenShiftException(String.format("Could not create resource %s in namespace '%s': %s", resource.getName(), namespace, e.getMessage()), e);
-		} catch (SocketTimeoutException e) {
-			throw new OpenShiftException(e, "Socket timeout creating resource %s", resource.getName());
+			
+		try {
+			Request.Builder builder = newRequestBuilderTo(endpoint.toString());
+			if(HttpMethod.GET != method) {
+				builder.method(method.toString(),
```
adietish: what about the other HTTP methods that dont have payloads like HEAD, GET, OPTIONS, TRACE, etc.?

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -200,122 +162,104 @@ public String getResourceURI(IResource resource) {
...
+			throw new OpenShiftException(e, "Unable to execute request to %s", endpoint);
 		}
 	}
-
+	
+	
 	@Override
-	public <T extends IResource> T create(String kind, String namespace, String name, String subresource, IResource payload) {
-		if(ResourceKind.LIST.equals(kind)) throw new UnsupportedOperationException("Generic create operation not supported for resource type 'List'");
+	public String ready() {
```
adietish: to me as noon ready() wont tell me exactly what this is doing. Unexpected that this does a healt check on the server, aka a remote call. I'd pretty much prefer a more expressive naming like #isServerReady() or similar

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -200,122 +162,104 @@ public String getResourceURI(IResource resource) {
...
-			LOGGER.debug(response);
-			return factory.create(response);
-		} catch (HttpClientException e){
-			throw createOpenShiftException(String.format("Could not create resource %s in namespace '%s': %s", resource.getName(), namespace, e.getMessage()), e);
-		} catch (SocketTimeoutException e) {
-			throw new OpenShiftException(e, "Socket timeout creating resource %s", resource.getName());
+			
+		try {
+			Request.Builder builder = newRequestBuilderTo(endpoint.toString());
+			if(HttpMethod.GET != method) {
```
adietish: why  not reusing #newRequestBuilder?

src/main/java/com/openshift/restclient/ClientBuilder.java
```diff
@@ -107,18 +140,41 @@ public ClientBuilder withConnectTimeout(int connectInMillis) {
...
+			OpenShiftAuthenticator authenticator = new OpenShiftAuthenticator();
+
+			OkHttpClient.Builder builder = new OkHttpClient.Builder()
+				.addInterceptor(responseCodeInterceptor)
+				.authenticator(authenticator)
+				.readTimeout(readTimeout, readTimeoutUnit)
+				.writeTimeout(writeTimeout, writeTimeoutUnit)
+				.connectTimeout(connectTimeout, connectTimeoutUnit)
+				.hostnameVerifier(this.sslCertificateCallback)
+				.followRedirects(false)
```
adietish: why not following redirects? quite frequently those allow endpoints to evolve while staying compatible (and other useful usecases)

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -508,15 +352,31 @@ public boolean equals(Object obj) {
...
-			return other.strategy == null;
+		if (authContext == null) {
+			return other.authContext == null;
 		} else {
-			if (other.strategy == null) {
+			if (other.authContext == null) {
 				return false;
 			}
-			return ObjectUtils.equals(strategy.getUsername(), other.strategy.getUsername());
+			return ObjectUtils.equals(authContext.getUserName(), other.authContext.getUserName());
```
adietish: Shouldnt this be impl'd in AuthContext#equals?

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -349,137 +293,37 @@ public  boolean supports(Class<? extends ICapability> capability) {
...
-	@Override
-	public IAuthorizationDetails getAuthorizationDetails(String baseURL) {
-		return this.authClient.getAuthorizationDetails(baseURL);
+	
+	public String getToken() {
+		return getAuthorizationContext().getToken();
 	}
 
 	@Override
-	public void setSSLCertificateCallback(ISSLCertificateCallback callback) {
```
adietish: dont we allow ssl certificate callbacks any more?

src/main/java/com/openshift/internal/restclient/authorization/AuthorizationContext.java
```diff
@@ -35,12 +46,34 @@ public AuthorizationContext(String token, String expires, IUser user, String sch
...
+		return context;
+	}
+	
+	public void setClient(IClient client){
+		this.client = client;
+	}
+	
 	@Override
 	public boolean isAuthorized() {
+		if(user == null) {
```
adietish: we're clearly changing the semantics here which is not a good idea imho. We now possibly do a remote call while before this wasnt the case.
I'd advocate this NOT to do any remote call. I'd let the methods that require a user to do the remote call.and create the user. What is user being used for? I dont see references?

src/main/java/com/openshift/internal/restclient/authorization/AuthorizationContext.java
```diff
@@ -35,12 +46,34 @@ public AuthorizationContext(String token, String expires, IUser user, String sch
...
+			synchronized (this) {
+				user = client.get(ResourceKind.USER, "~", "");
```
adietish: we're clearly changing the semantics here which is not a good idea imho. We now possibly do a remote call while before this wasnt the case.
I'd advocate this NOT to do any remote call. I'd let the methods that require a user to do the remote call.and create the user. What is user being used for? I dont see references?

src/main/java/com/openshift/internal/restclient/http/BadRequestException.java
```diff
@@ -1,27 +0,0 @@
...
- * 
- * Contributors: 
- * Red Hat, Inc. - initial API and implementation 
- ******************************************************************************/ 
-package com.openshift.internal.restclient.http;
-
-/**
- * @author André Dietisheim
- */
-public class BadRequestException extends HttpClientException {
```
adietish: We now expose OkHttp exceptions?

src/main/java/com/openshift/internal/restclient/okhttp/BasicChallangeHandler.java
```diff
@@ -0,0 +1,42 @@
...
+import okhttp3.Credentials;
+import okhttp3.Headers;
+import okhttp3.Request.Builder;
+
+/**
+ * 
+ * @author jeff.cantrill
+ *
+ */
+public class BasicChallangeHandler implements IChallangeHandler{
```
adietish: dont we support OAuth yet? (I dont see any other IChallangeHandler than the BasiChallengeHandler)

src/main/java/com/openshift/internal/restclient/okhttp/OpenShiftAuthenticator.java
```diff
@@ -0,0 +1,139 @@
...
+	private Collection<IChallangeHandler> challangeHandlers = new ArrayList<>();
+	private OkHttpClient okClient;
+	private IClient client;
+	
+	@Override
+	public Request authenticate(Route route, Response response) throws IOException {
+		if(unauthorizedForCluster(response)){
+			String requestUrl = response.request().url().toString();
+			Request authRequest = new Request.Builder()
+					.addHeader("X-CSRF-Token", "1")
```
adietish: want to extract used header keys and values to a constants class for improved readability?

src/main/java/com/openshift/internal/restclient/okhttp/OpenShiftAuthenticator.java
```diff
@@ -0,0 +1,139 @@
...
+		.newBuilder()
+		.authenticator(new Authenticator() {
+			
+			@Override
+			public Request authenticate(Route route, Response response) throws IOException {
+				if(StringUtils.isNotBlank(response.request().header(AUTH_ATTEMPTS))) {
+					return null;
+				}
+				if(StringUtils.isNotBlank(response.header(IHttpConstants.PROPERTY_WWW_AUTHENTICATE))) {
+					for (IChallangeHandler challangeHandler : challangeHandlers) {
```
adietish: want to extract iterating over handlers and checking if they can handle the required auth method to some metod called #getChallengeHandler(requestHeaders)?

src/main/java/com/openshift/internal/restclient/okhttp/OpenShiftAuthenticator.java
```diff
@@ -0,0 +1,139 @@
...
+	private IAuthorizationDetails captureAuthDetails(String url) {
+		IAuthorizationDetails details = null;
+		Map<String, String> pairs = URIUtils.splitFragment(url);
+		if (pairs.containsKey(ERROR)) {
+			details = new AuthorizationDetails(pairs.get(ERROR), pairs.get(ERROR_DETAILS));
+		}
+		return details;
+	}
+	
+	private String captureToken(Response response) {
```
adietish: to me the method name wont tell me that this is extracting the token from the given response and set it to the authContext.
I'd prefer something more explicit like #setAuthContextToken(Response)

src/main/java/com/openshift/internal/restclient/okhttp/OpenShiftAuthenticator.java
```diff
@@ -0,0 +1,139 @@
...
+		return null;
+	}
+	
+	private boolean unauthorizedForCluster(Response response) {
+		String requestHost = response.request().url().host();
+		return response.code() == IHttpConstants.STATUS_UNAUTHORIZED && client.getBaseURL().getHost().equals(requestHost);
+	}
+	
+	private Response tryAuth(Request authRequest) throws IOException {
+		return okClient
```
adietish: what if user forgot to call #setOkClient and okClient therefore is null here?

src/main/java/com/openshift/internal/restclient/okhttp/OpenShiftAuthenticator.java
```diff
@@ -0,0 +1,139 @@
...
+		return details;
+	}
+	
+	private String captureToken(Response response) {
+		String token = null;
+		Map<String, String> pairs = URIUtils.splitFragment(response.header(PROPERTY_LOCATION));
+		if (pairs.containsKey(ACCESS_TOKEN)) {
+			token = pairs.get(ACCESS_TOKEN);
+			IAuthorizationContext authContext = client.getAuthorizationContext();
+			if(authContext != null) {
```
adietish: shouldnt we let the user know that we couldnt find the context? What's the situation in that case? Is it an error, does it mean than auth wont be possible or ...?

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -145,23 +121,9 @@ public String getResourceURI(IResource resource) {
...
-			String response = client.get(endpoint,  IHttpClient.DEFAULT_READ_TIMEOUT);
-			LOGGER.debug(String.format("List Response: %s:", response));
-			List<T> items = (List<T>) factory.createList(response, kind);
-			return filterItems(items, labels); //client filter until we can figure out how to restrict with a server call
-		} catch (HttpClientException e){
-			throw createOpenShiftException(String.format("Could not list %s resources in namespace %s: %s", kind, namespace, e.getMessage()), e);
-		} catch (SocketTimeoutException e) {
-			throw new OpenShiftException(e, "Socket timeout listing resources");
-		} 
+		IList list = get(kind, namespace);
```
jcantrill: TypeMapper is used by the UrlBuilder and checked for supported types in UrlBuilder#buildWithNamespaceInPath and will throw an UnsupportedEndpointException

jcantrill: add throws to the documentation

src/main/java/com/openshift/internal/restclient/DefaultClient.java
jcantrill: Updated class to have new method

src/main/java/com/openshift/internal/restclient/DefaultClient.java
jcantrill: Removed line as the UrlBuilder along with the TypeMapper for apigroups handles this case

src/main/java/com/openshift/internal/restclient/DefaultClient.java
jcantrill: I have not encountered this in the context of OpenShift so they were not needed.  Should I add them here? Due you have a complete list?

src/main/java/com/openshift/restclient/authorization/IAuthorizationContext.java
```diff
@@ -26,7 +26,7 @@
 	
 	/**
 	 * 
-	 * @return true if the token is non-null; false otherwise
+	 * @return true if authorized; false otherwise
```
adietish: the impl now seems to do remote calls which should at least be mentioned here. But I believe that we should not do remote calls in this metod since it's a change in semantics which I'd only do if really really needed..

src/main/java/com/openshift/internal/restclient/DefaultClient.java
jcantrill: I dont understand what you are saying here.

src/main/java/com/openshift/restclient/model/template/ITemplate.java
```diff
@@ -25,9 +25,17 @@
 	 * creates
 	 * @return
 	 */
+	@Deprecated
```
adietish: I'd add informations in the javadoc what the user should now do instead.

src/main/java/com/openshift/internal/restclient/DefaultClient.java
jcantrill: changed to getServerReadyStatus() since it returns a string to be evaluated.

src/main/java/com/openshift/restclient/model/template/ITemplate.java
```diff
@@ -25,9 +25,17 @@
...
 	Collection<IResource> getItems();
 	
 	/**
+	 * Retrieves the list of resources this template
+	 * creates
+	 * @return
+	 */
+	Collection<IResource> getObjects();
```
adietish: better call it #getResources since it returns resources?

src/main/java/com/openshift/internal/restclient/DefaultClient.java
jcantrill: We do but through the initialization of the client via the ClientBuilder.  I couldn't find a case where it needed to exist as a setter on the client.  In the JBT code we set the callback when we are creating new connections which rely on ClientBuilder

src/main/java/com/openshift/internal/restclient/authorization/AuthorizationContext.java
jcantrill: That is not strictly true as before we assumed you were authorized if your token was not null which is not necessarily correct; it just means you have a token.  In JBT we had logic that more or less did what was doing here so you were making a remote call anyway.  Since we simplified how the auth logic is consumed, I believe this is a reasonable trade off. At some point or another, remote calls are going to need to be made in the authorization process.

src/main/java/com/openshift/internal/restclient/http/BadRequestException.java
jcantrill: We should not be exposing those exception.  I believe I have captured all exceptions and wrappered them in our OpenShiftException or some flavor of that.

src/main/java/com/openshift/internal/restclient/okhttp/BasicChallangeHandler.java
jcantrill: Challange handlers are only for auth that returns www-authenticate header.  To my knowledge, there is no way for us to progammatically handle oauth; its the same way with 'oc login'  We have a pretty good story now with jbt which allows us to navigate to a browser which is essentially what 'oc login' tells you; go to THIS url to get your token and give it too me.

src/main/java/com/openshift/internal/restclient/okhttp/OpenShiftAuthenticator.java
jcantrill: Users should only be using the ClientBuilder to crate clients which is the only user of this class; this class not intended to be used publicly.  If we have that situation then either we have a problem (bug) or someone is using it and they should not be.  I know not the best answer, but....

src/main/java/com/openshift/internal/restclient/okhttp/OpenShiftAuthenticator.java
```diff
@@ -0,0 +1,139 @@
...
+				if(StringUtils.isNotBlank(response.header(IHttpConstants.PROPERTY_WWW_AUTHENTICATE))) {
+					for (IChallangeHandler challangeHandler : challangeHandlers) {
+						if(!challangeHandler.canHandle(response.headers())) {
+							Builder requestBuilder = response.request().newBuilder()
+									.header(AUTH_ATTEMPTS, "1");
+							return challangeHandler.handleChallange(requestBuilder).build();
+						}
+					}
+				}
+				return null;
```
jcantrill: No, this is the indicator to stop calling authenticators per the lib documentation.

src/main/java/com/openshift/internal/restclient/okhttp/OpenShiftAuthenticator.java
jcantrill: renamed to extactAndSetAuthContextToken

src/main/java/com/openshift/internal/restclient/okhttp/OpenShiftAuthenticator.java
jcantrill: The authContext should never be null since I made a continuous decision for it to never be null.

src/main/java/com/openshift/restclient/ClientBuilder.java
jcantrill: Re-tested to prove how we this could be made 'true' and updated..  Updated on auth logict to not follow redirects

src/main/java/com/openshift/restclient/model/template/ITemplate.java
jcantrill: This is what its called in the kube json model

jcantrill:
[merge]

PR #181, opened 2016-06-28
Add a volume source interface

lulf:
This pull request adds an interface to the replicationcontroller for adding volumes to a pod spec. I have only implemented support for EmptyDir, Secret and PersistentVolumeClaim, as that is what we need right now. Adding other implementations should be straightforward.

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/model/volume/EmptyDirVolumeSource.java
+++ b/src/main/java/com/openshift/internal/restclient/model/volume/EmptyDirVolumeSource.java
@@ -1,0 +1,47 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package com.openshift.internal.restclient.model.volume;
+
+import com.openshift.restclient.model.volume.IEmptyDirVolumeSource;
+import com.openshift.restclient.model.volume.VolumeType;
+import org.jboss.dmr.ModelNode;
+
+import static com.openshift.internal.util.JBossDmrExtentions.asString;
+import static com.openshift.internal.util.JBossDmrExtentions.set;
+
+/**
+ * @author Ulf Lilleengen
+ */
+public class EmptyDirVolumeSource extends VolumeSource implements IEmptyDirVolumeSource {
+    private static final String PROP_MEDIUM = "medium";
+
+    private final ModelNode node;
+
+    public EmptyDirVolumeSource(ModelNode node) {
+        super(node);
+        this.node = node.get(VolumeType.EMPTY_DIR);
+    }
+
+    public EmptyDirVolumeSource(String name) {
+        this(new ModelNode());
+        setName(name);
+    }
+
+    @Override
+    public String getMedium() {
+        return asString(node, getPropertyKeys(), PROP_MEDIUM);
+    }
+
+    @Override
+    public void setMedium(String medium) {
+        set(node, getPropertyKeys(), PROP_MEDIUM, medium);
+    }
+}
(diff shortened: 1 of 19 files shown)


src/main/java/com/openshift/internal/restclient/model/volume/EmptyDirVolumeSource.java
```diff
@@ -0,0 +1,37 @@
+package com.openshift.internal.restclient.model.volume;
```
jcantrill: please copy and add the copyright and licensing info

src/main/java/com/openshift/internal/restclient/model/volume/PersistentVolumeClaimVolumeSource.java
```diff
@@ -0,0 +1,52 @@
...
+        return asString(node, getPropertyKeys(), PROP_CLAIM_NAME);
+    }
+
+    @Override
+    public void setClaimName(String claimName) {
+        set(node, getPropertyKeys(), PROP_CLAIM_NAME, claimName);
+    }
+
+    @Override
+    public boolean getReadOnly() {
```
jcantrill: isReadOnly would be more convention

src/main/java/com/openshift/internal/restclient/model/volume/VolumeSource.java
```diff
@@ -39,4 +39,35 @@ public String getName() {
...
 	}
 
+    @Override
+    public void setName(String name) {
+        set(getNode(), getPropertyKeys(), NAME, name);
+    }
+
+    @Override
+    public String toJSONString() {
+        if (!get(getNode(), getPropertyKeys(), NAME).isDefined()) {
```
jcantrill: You can check with something like StringUtils.isBlank(getName()) since all the getters will return a value, specifically in this case it will return an emptystring which is bad given name is essentially the identity.

src/main/java/com/openshift/internal/restclient/model/volume/VolumeSource.java
```diff
@@ -14,22 +14,22 @@
...
-import com.openshift.restclient.model.volume.IVolume;
 import com.openshift.restclient.model.volume.IVolumeSource;
 
 /**
  * 
  * @author Jeff Cantrill
  *
  */
-public class VolumeSource 
+public abstract class VolumeSource
```
jcantrill: we should consider modifying this class to extend from KubernetesResource instead of the ModelNodeAdapter since it looks like we can do REST operations on persistentvolumeclaims

src/main/java/com/openshift/internal/restclient/model/volume/VolumeSource.java
```diff
@@ -39,4 +39,35 @@ public String getName() {
...
+
+    public static IVolumeSource create(ModelNode node) {
+        if (node.has(VolumeType.EMPTY_DIR)) {
+            return new EmptyDirVolumeSource(node);
+        } else if (node.has(VolumeType.SECRET)) {
+            return new SecretVolumeSource(node);
+        } else if (node.has(VolumeType.PERSISTENT_VOLUME_CLAIM)) {
+            return new PersistentVolumeClaimVolumeSource(node);
+        } else {
+            return new UnsupportedVolumeSource(node);
```
jcantrill: We should probably return an anonymous inner class implementation here instead of 'Unsupported' since we are returning something.  Unsupported makes it sound like we are unable to do anything without though its really some generic representation of a volume source since we have no specifics at this time.

src/main/java/com/openshift/restclient/model/IReplicationController.java
```diff
@@ -174,5 +174,12 @@
 	 */
 	Set<IVolumeSource> getVolumes();
 
+	/**
```
jcantrill: Maybe also note that volumes with duplicate names will not be added.

src/main/java/com/openshift/internal/restclient/model/volume/VolumeSource.java
jcantrill: This change will depend on the diff between whats in this object and whats returned from the REST operations.  i would hope they are the same but if not then please ignore this comment until we can further evaluate it.

lulf: The volume sources has no metadata, kind or anything like that like a resource has. Its not defined as a resource by the REST api, see https://docs.openshift.org/latest/rest_api/kubernetes_v1.html#v1-volume and https://docs.openshift.org/latest/rest_api/kubernetes_v1.html#v1-emptydirvolumesource for instance.

There is PersistentVolumeClaim resource, but that should be a different interface, and already exists in the form of IPersistentVolumeClaim. This is more like a reference to that resource.

jcantrill:
[merge]

PR #187, opened 2016-07-08
[OSJC-263] fix unable to delete OpenShift resources

jcantrill:
[test]

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/DefaultClient.java
+++ b/src/main/java/com/openshift/internal/restclient/DefaultClient.java
@@ -193,14 +193,9 @@ public class DefaultClient implements IClient, IHttpConstants{
 				.build();
 			
 		try {
-			Request.Builder builder = newRequestBuilderTo(endpoint.toString());
-			if(!HttpMethod.GET.toString().equals(method)) {
-				String json = payload == null ? "" : payload.toJson(true);
-				builder.method(method.toString(), 
-						RequestBody.create(MediaType.parse(MEDIATYPE_APPLICATION_JSON), json));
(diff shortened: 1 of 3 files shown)


src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -212,6 +207,18 @@ public String getResourceURI(IResource resource) {
 		}
 	}
 	
+	private RequestBody getPayload(String method, IResource payload) {
```
jcantrill: @adietish this fixes my issues, but I'm wondering if it should evaluate based on the payload. I *THINK* delete may allow you to provide a payload which would allow you to define delete options (e.g. grace period), but I have not further investigated.  Do you think this is good for now? Or should we throw in cases like create (POST) where you should provide a payload not allow the library to send empty?

jcantrill: cc @scabanovich I would be interested in your thoughts regarding my previous comment

jcantrill: resolved my own question.  Open issue https://issues.jboss.org/browse/OSJC-249 that resolves

adietish:
@jcantrill afaics https://www.ietf.org/rfc/rfc2616.txt wont explicitly state if a body is required for DELETE (and POST). I'd therefore prefer the lenient approach where we only fail if the server tells us that the request could not be properly handled.

jcantrill:
[merge]ing as we can fix if needed.  @adietish I think that adheres to what you are saying

PR #199, opened 2016-08-09
setter for proxyAuthenticator (okhttp)

kaimago:
setter for ProxyAuthenticatior is needed


src/main/java/com/openshift/restclient/ClientBuilder.java
```diff
@@ -33,6 +33,7 @@
 import com.openshift.restclient.http.IHttpConstants;
 import com.openshift.restclient.utils.SSLUtils;
 
+import okhttp3.Authenticator;
```
jcantrill: @fbricon @adietish This would be the first place where we publicly expose okhttp to consumers of this client library.  I would prefer we had some sort of abstraction though I don't know how practical it is to do so.  Thoughts on allowing the authenticator to go public?  I have reservations.

fbricon: Yeah I'm not found about having to expose the okhttp API. +1 on having an abstraction

adietish: +1, I dont think that it's a good idea to expose the client library that we use. This would disallows us from moving off it without breaking API compat.
Thinking about an abstraction, I looked up a few of the most widely used http client libs. Shouldnt be too hard to provide an abstraction that's capable of using the following libs behind the scenes:
- okHttp: https://square.github.io/okhttp/3.x/okhttp/okhttp3/Authenticator.html
- apache httpclient: http://hc.apache.org/httpclient-3.x/apidocs/org/apache/commons/httpclient/auth/AuthScheme.html
- jdk HttpUrlConnection: https://docs.oracle.com/javase/7/docs/api/java/net/Authenticator.html
- etc.

kaimago: any progress for the abstraction ? 

PS: It's not a big deal to change the authenticator, when okhttp lib will really change ;)

jcantrill:
@kaimago We are unable to justify the effort to properly wrapper the proxy authenticator.  Please rebase and update your PR so we can merge it into the code base.

kaimago:
Sorry PR is a mess. I will make a new one.

PR #204, opened 2016-08-15
Add support for exec type lifecycle hooks

lulf:


src/main/java/com/openshift/internal/restclient/model/Container.java
```diff
@@ -142,13 +143,19 @@ public String getImagePullPolicy() {
...
+	public void setLifecycle(ILifecycle lifecycle) {
+		ModelNode lifecycleNode = ModelNode.fromJSONString(lifecycle.toJson());
+		get(node, propertyKeys, LIFECYCLE).set(lifecycleNode);
 	}
 
 	@Override
-	public String getLifecycle() {
-		return asString(node, propertyKeys, LIFECYCLE);
+	public ILifecycle getLifecycle() {
+		ModelNode lifecycle = get(node, propertyKeys, LIFECYCLE);
```
jcantrill: I know elsewhere we use this pattern, but recently introduced 'has' for checking if its there, which is what we should use.  The benefit is that it wont add the node to the underlying json if its 'not defined'

src/main/java/com/openshift/internal/restclient/model/Lifecycle.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.Collections;
+import java.util.Optional;
+
+import static com.openshift.internal.util.JBossDmrExtentions.get;
+
+/**
+ * @author Ulf Lilleengen
+ */
+public class Lifecycle implements ILifecycle {
```
jcantrill: I would consider extending ModelNodeAdapter which is like a light-weight KubernetesResource that allows you to contain the ModelNode.  This will allow you to remove fromJson and simply clone or provide an alternative Constructor that takes a ModelNode

lulf:
@jcantrill I've updated the PR with some changes after our discussion:

* Remove extra indirection that allowed possible misuse of the API by setting both exec and http actions, which is disallowed
* Use builders for building lifecycle and commands for exec.
* Make lifecycles read-only once built, due to the REST API doc stating "Cannot be updated."

I think the API is fairly easy to use correctly now. 

Re: ModelNodeAdapter, I think the nice property right of the code right now is that all serialization/deserialization code is in self-contained methods rather than embedded in the class. If you insist, I can change it, but IMO it would be even better to keep model and serialization format as separate as possible.

src/main/java/com/openshift/internal/restclient/model/Container.java
```diff
@@ -142,13 +143,18 @@ public String getImagePullPolicy() {
...
+		get(node, propertyKeys, LIFECYCLE).set(lifecycleNode);
 	}
 
 	@Override
-	public String getLifecycle() {
-		return asString(node, propertyKeys, LIFECYCLE);
+	public ILifecycle getLifecycle() {
+		if (node.has(LIFECYCLE)) {
+			return Lifecycle.fromJson(get(node, propertyKeys, LIFECYCLE));
+		} else {
```
jcantrill: Are we certain we always wish to return something and not null?  I'm not suggesting its wrong, just asking the quesion

src/main/java/com/openshift/internal/restclient/model/ExecAction.java
```diff
@@ -0,0 +1,66 @@
+package com.openshift.internal.restclient.model;
```
jcantrill: License and copyright block

src/main/java/com/openshift/internal/restclient/model/Lifecycle.java
```diff
@@ -0,0 +1,82 @@
+package com.openshift.internal.restclient.model;
```
jcantrill: License and copyright block

src/main/java/com/openshift/restclient/model/IExecAction.java
```diff
@@ -0,0 +1,13 @@
+package com.openshift.restclient.model;
```
jcantrill: License and copyright block

src/main/java/com/openshift/restclient/model/IHandler.java
```diff
@@ -0,0 +1,27 @@
+package com.openshift.restclient.model;
```
jcantrill: License and copyright block

src/main/java/com/openshift/restclient/model/IHandler.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.openshift.internal.restclient.model.JSONSerializeable;
+
+import java.util.Optional;
+
+/**
+ * @author Ulf Lilleengen
+ */
+public interface IHandler extends JSONSerializeable {
+    enum HandlerType {
```
jcantrill: We have made a deliberate decision to avoid proper enums  in order to make the library lenient to changes.  Please convert to const.

src/main/java/com/openshift/restclient/model/ILifecycle.java
```diff
@@ -0,0 +1,19 @@
+package com.openshift.restclient.model;
```
jcantrill: License and copyright block

src/test/java/com/openshift/internal/restclient/model/v1/LifecycleTest.java
```diff
@@ -0,0 +1,70 @@
+package com.openshift.internal.restclient.model.v1;
```
jcantrill: License and copyright block

src/main/java/com/openshift/internal/restclient/model/Container.java
lulf: Ideally, I'd return Optional.empty, but that would require changing the API. Now it will return an 'empty' lifecycle, which is not ideal. Perhaps null is better.

jcantrill:
[merge]

PR #205, opened 2016-08-24
[JBIDE-23030] Add dc#scale capability

jcantrill:
@fbricon @adietish 

Adds the scale capability and new package structure to try and mirror organization of the server

[test]

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/apis/TypeMetaFactory.java
+++ b/src/main/java/com/openshift/internal/restclient/apis/TypeMetaFactory.java
@@ -1,0 +1,106 @@
+/******************************************************************************* 
+ * Copyright (c) 2016 Red Hat, Inc. 
+ * Distributed under license by Red Hat, Inc. All rights reserved. 
+ * This program is made available under the terms of the 
+ * Eclipse Public License v1.0 which accompanies this distribution, 
+ * and is available at http://www.eclipse.org/legal/epl-v10.html 
+ * 
+ * Contributors: 
+ * Red Hat, Inc. - initial API and implementation 
+ ******************************************************************************/
+package com.openshift.internal.restclient.apis;
+
+import java.lang.reflect.Constructor;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+import org.apache.commons.lang.StringUtils;
+import org.jboss.dmr.ModelNode;
+
+import com.openshift.internal.restclient.api.models.TypeMeta;
+import com.openshift.internal.restclient.apis.autoscaling.models.Scale;
+import com.openshift.internal.restclient.model.properties.ResourcePropertiesRegistry;
+import com.openshift.internal.restclient.model.properties.ResourcePropertyKeys;
+import com.openshift.internal.util.JBossDmrExtentions;
+import com.openshift.restclient.OpenShiftException;
+import com.openshift.restclient.ResourceFactoryException;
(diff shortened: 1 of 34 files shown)


jcantrill:
[test]

src/main/java/com/openshift/internal/restclient/apis/TypeMetaFactory.java
```diff
@@ -0,0 +1,106 @@
...
+import com.openshift.internal.util.JBossDmrExtentions;
+import com.openshift.restclient.OpenShiftException;
+import com.openshift.restclient.ResourceFactoryException;
+import com.openshift.restclient.UnsupportedVersionException;
+import com.openshift.restclient.api.ITypeFactory;
+import com.openshift.restclient.api.models.INameSetable;
+import com.openshift.restclient.api.models.INamespaceSetable;
+import com.openshift.restclient.api.models.ITypeMeta;
+
+public class TypeMetaFactory implements ITypeFactory, ResourcePropertyKeys {
```
adietish: what's the difference of this to ResourceFactory?

src/main/java/com/openshift/internal/restclient/apis/TypeMetaFactory.java
```diff
@@ -0,0 +1,106 @@
...
+	
+	private static final String DELIMITER = ".";
+	private static final Map<String, Class<? extends ITypeMeta>> IMPL_MAP = new HashMap<>();
+
+	static {
+		IMPL_MAP.put("Scale", Scale.class);
+	}
+	
+	@Override
+	public Object stubKind(String kind, Optional<String> name, Optional<String> namespace) {
```
adietish: in sake of a more uniformous API I would keep using #stub like in ResourceFactory.

src/main/java/com/openshift/internal/restclient/apis/TypeMetaFactory.java
jcantrill: Of course its been a while since I looked at this, but I believe I introduced a new factory to handle cases for kinds that are not necessarily resources that are capable of being submitted to the server.  In this specific case, IIR, you can scale a DC which returns a 'scale' kind, but the kind is not what we have been calling a 'resource' and is not capable of submission to the server.  It's more of a struct for either containerizing parameters or results.  This was specifically introduced to 'reuse' some logic

src/main/java/com/openshift/internal/restclient/apis/TypeMetaFactory.java
jcantrill: I like 'stub' better but I don't think I was able to reuse the method name because of the return type

adietish: I would still stick to uniformous names, even if not technically overriding over overloading. It makes using an API easier.

adietish:
other than these minor LGTM

src/main/java/com/openshift/internal/restclient/apis/TypeMetaFactory.java
jcantrill: I agree with you but not certain how overcome the compiler error

adietish:
@jcantrill any updates on this? can we merge this?

jcantrill:
@adietish I have no updates as long as you are good with it

jcantrill:
[merge] as @adietish LGTM'd it

PR #207, opened 2016-09-09
[JBIDE-23015] - Creating a route should have a default port

jeffmaury:
- Extends Route model

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/model/Route.java
+++ b/src/main/java/com/openshift/internal/restclient/model/Route.java
@@ -16,6 +16,7 @@ import com.openshift.restclient.IClient;
 import com.openshift.restclient.ResourceKind;
 import com.openshift.restclient.model.route.IRoute;
 import com.openshift.restclient.model.route.ITLSConfig;
+import com.openshift.restclient.model.route.ITargetPort;
 
 /**
  * @author Jeff Cantrill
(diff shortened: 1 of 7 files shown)


src/main/java/com/openshift/internal/restclient/model/Route.java
```diff
@@ -152,4 +173,31 @@ public void setDestinationCertificate(String destinationCertificate) {
 		}
 
 	}
+	
+	private class TargetPort implements ITargetPort {
+
+	    private boolean isString = true;
```
adietish: "isString" seems unused?

src/test/java/com/openshift/internal/restclient/model/v1/RouteTest.java
```diff
@@ -63,7 +64,42 @@ public void createTLSConfigWhenUndefined() {
 		assertEquals("", tls.getKey());
 	}
 	
-	@Test
+    @Test
+    public void getPortWhenUndefined() {
+        ModelNode node = ModelNode.fromJSONString(Samples.V1_ROUTE_WO_TLS.getContentAsString());
+        route = new Route(node, client, ResourcePropertiesRegistry.getInstance().get(VERSION, ResourceKind.ROUTE));
+        ITargetPort port = route.getPort();
+        assertNull(port);
```
adietish: Shouldn't we, in an additional test, also assert that the port, that's returned if you route.getPort().getTargetPort() is -1?

jeffmaury: Do you mean if you don't select a particular port. In that case, getPort() will return null

adietish: I see a Route#createPort(). If nothing further is done (which is most likely in user code, not included in this PR) I see a TargetPort being create that will return -1 if you dont set it a numeric port. Missing something?

adietish:
+1, LGTM

adietish:
[merge]

Issue #215, opened 2016-09-23
List is returning Objects instead of Items


transamericamoon:
I can't find a spec for Kind: List but every example i've seen it returns items.  For some reason yours returns Objects.  From what i've seen only templates have Objects.  If this is incorrect can you point me to the API specs for kind: List.

Issue is here:
https://github.com/openshift/openshift-restclient-java/blob/master/src/main/java/com/openshift/internal/restclient/model/List.java#L74

I can submit a PR if that will help.

PR #216, opened 2016-09-26
Fix List to produce items instead of objects

transamericamoon:
From what i've seen List contains items and templates contain objects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ Testing
 -------
 
 To run the integration tests:
+
 1.  Define a user with cluster admin privilege
 1.  Download the oc binary
 1.  Run the tests: `mvn integration-test -Pintegration-tests -DserverURL=https://localhost:8443 -Ddefault.cluster.admin=foo -Ddefault.cluster.password=bar -Ddefault.openshift.location=/tmp/oc`


jcantrill:
@transamericamoon I am unable to find the reason why 'objects' was introduced.  The only thing I can say/recall is that it was introduced during a refactoring in which something was believed to be true and apparently was not.  Looking at the api for List, I agree that I do not see any reference to 'object'.  Could you add https://github.com/openshift/openshift-restclient-java/pull/216/files#diff-62e85ab15c1ef21309b9638e77e8790cR44 to your fix and squash your commits?

jcantrill:
This will fix #215

jeffmaury:
Unit tests ?

jcantrill:
- ListTest

 already https://github.com/openshift/openshift-restclient-java/blob/master/src/test/java/com/openshift/internal/restclient/model/v1/ListTest.java  
- please see note regarding also removing Objects from: https://github.com/openshift/openshift-restclient-java/pull/216/files#diff-62e85ab15c1ef21309b9638e77e8790cR44
- please squash your commits

jeffmaury:
Yes but the test doesn't test the fixed feature/behavior

transamericamoon:
@jeffmaury Not sure how to test it since i don't have access to the ModelNode in the test.  I serialize and check for "items" previously it was creating "objects".

src/main/java/com/openshift/internal/restclient/model/List.java
```diff
@@ -41,10 +41,9 @@ public List(ModelNode node, IClient client, Map<String, String []> propertyKeys)
 	@Override
 	public Collection<IResource> getItems(){
 		if(items == null) {
-			String key = getNode().has(OBJECTS) ? OBJECTS : ITEMS;
-			ModelNode listNode = get(key);
+			ModelNode listNode = get(ITEMS);
 			if (listNode.isDefined()) {
-				Collection<ModelNode> nodes = get(key).asList();
+				Collection<ModelNode> nodes = get(ITEMS).asList();
```
jcantrill: listNode.asList()

jcantrill:
@transamericamoon the only way to get access to the node is via casting to a KubernetesResource and calling the getNode methods.

transamericamoon:
Made the suggested changes.  Thanks for the guidance.

transamericamoon:
I just realized I used spaces instead of tabs in the unit test, sorry :(

Issue #214, opened 2016-09-22
Support Jolokia request


kaimago:
To Support Jolokia-Request like this:

https://baseurl/api/v1/namespaces/projectname/pods/https:podname-7-fl9hk:8778/proxy/jolokia/exec/java.util.logging:type=Logging/getLoggerLevel/org

execute Methode just need an additional parameter to put e.g. "/jolokia/exec/java.util.logging:type=Logging/getLoggerLevel/org" add the end of the request.

execute(String httpMethod, String kind, String namespace, String name, String subresource, String subcontext, IResource payload);

Should I make a PR or is this a function you don't want to support ?

jcantrill:
Feel free to submit a PR that overloads that function.  I would suggest it either takes a map or a vararg for additional parameters.  Or what might be even better, I'm wondering, if it would be better to overload with a parameter that is some sort of callback.  Callback would provide additional processing since the URL pattern does not appear to be query args, but something more; it appends to the path.  @adietish thoughts?

jeffmaury:
+1 for being a little more generic, thus adding a map or a callback (as I don't know if this proxy kind of request is being standardized or not)

PR #218, opened 2016-09-29
addSubContext to UrlBuilder

kaimago:
Adding SubContext to execute jolokia request.

https://github.com/openshift/openshift-restclient-java/issues/214#issuecomment-249626878
Adding a map is not useful in my opinion, because it's just a subContext that is needed and not additional parameters.


jcantrill:
[merge]

PR #220, opened 2016-09-30
Adding pod exec function

jupierce:
Feature to support exec function for OpenShift Jenkins plugin: https://trello.com/c/Yl4qomsw/878-5-additional-jenkins-plugin-enhancement-requests-from-github

fyi @gabemontero 

API use example:

```
            final IPodExec.Options options = new IPodExec.Options();
            options.stdErr(true);
            options.stdOut(true);
            options.container("ruby-hello-world");


            IPod pod = client.get("pods", "ruby-hello-world-1-hy2xq", "myproject");
            IStoppable v = pod.accept(new CapabilityVisitor<IPodExec, IStoppable>() {
                @Override
                public IStoppable visit(IPodExec capability) {
                    return capability.start( new IPodExec.IPodExecOutputListener() {
                        @Override
                        public void onOpen() {
                            System.out.println( "OPEN!" );
                        }

                        @Override
                        public void onStdOut(String message) {
                            System.out.println( "stdout: " + message );
                        }

                        @Override
                        public void onStdErr(String message) {
                            System.out.println( "stderr: " + message );
                        }

                        @Override
                        public void onExecErr(String message) {
                            System.out.println( "execerr: " + message );
                        }

                        @Override
                        public void onClose(int code, String reason) {
                            System.out.println( "CLOSE! [" + code + "]: " + reason  );
                        }

                        @Override
                        public void onFailure(IOException e) {
                            e.printStackTrace();
                        }

                    }, options, commands );
                }
            }, null);
```

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/DefaultClient.java
+++ b/src/main/java/com/openshift/internal/restclient/DefaultClient.java
@@ -253,10 +253,14 @@ public class DefaultClient implements IClient, IHttpConstants{
 	}
 
 	public Request.Builder newRequestBuilderTo(String endpoint){
+		return newRequestBuilderTo(endpoint, MEDIATYPE_APPLICATION_JSON);
+	}
+
+	public Request.Builder newRequestBuilderTo(String endpoint,String acceptMediaType){
 		Request.Builder builder = new Request.Builder()
-			.url(endpoint.toString())
-			.header(PROPERTY_ACCEPT, MEDIATYPE_APPLICATION_JSON);
-		
+				.url(endpoint.toString())
+				.header(PROPERTY_ACCEPT, acceptMediaType);
+
 		String token =  null;
 		if(this.authContext != null &&  StringUtils.isNotBlank(this.authContext.getToken())){
 			token = this.authContext.getToken();
@@ -264,7 +268,8 @@ public class DefaultClient implements IClient, IHttpConstants{
 		builder.header(IHttpConstants.PROPERTY_AUTHORIZATION, String.format("%s %s", IHttpConstants.AUTHORIZATION_BEARER, token));
 		return builder;
 	}
-	
+
+
 	@Override
 	public <T extends IResource> T update(T resource) {
 		return execute(HttpMethod.PUT, resource.getKind(), resource.getNamespace(), resource.getName(), null, resource);
--- a/src/main/java/com/openshift/internal/restclient/URLBuilder.java
+++ b/src/main/java/com/openshift/internal/restclient/URLBuilder.java
@@ -8,19 +8,6 @@
  ******************************************************************************/
 package com.openshift.internal.restclient;
 
-import java.io.UnsupportedEncodingException;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.net.URLEncoder;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.Map;
-import java.util.Map.Entry;
-
-import org.apache.commons.lang.StringUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import com.openshift.restclient.IApiTypeMapper;
 import com.openshift.restclient.IApiTypeMapper.IVersionedApiResource;
 import com.openshift.restclient.OpenShiftException;
@@ -28,6 +15,18 @@ import com.openshift.restclient.ResourceKind;
 import com.openshift.restclient.UnsupportedEndpointException;
 import com.openshift.restclient.http.IHttpConstants;
 import com.openshift.restclient.model.IResource;
+import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLEncoder;
+import java.util.AbstractMap;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.Map;
 
 /**
  * Helper class to build the URL connection string in the proper
@@ -42,7 +41,7 @@ public class URLBuilder {
 	private String baseUrl;
 	private String kind;
 	private String name;
-	private Map<String, String> params = new HashMap<String, String>();
+	private ArrayList<AbstractMap.SimpleEntry<String,String>> params = new ArrayList<>();
 	private final IApiTypeMapper typeMappings;
 
 	private String apiVersion;
@@ -105,7 +104,7 @@ public class URLBuilder {
 	}
 
 	public URLBuilder addParmeter(String key, String value) {
-		params.put(key, value);
+		params.add(new AbstractMap.SimpleEntry<>( key, value ));
 		return this;
 	}
 	
@@ -175,10 +174,8 @@ public class URLBuilder {
 	private StringBuilder appendParameters(StringBuilder url) {
 		if (!params.isEmpty()) {
 			url.append(IHttpConstants.QUESTION_MARK);
-			for (Iterator<Entry<String, String>> iterator = params.entrySet()
-					.iterator(); iterator.hasNext();) {
-				Entry<String, String> entry = (Entry<String, String>) iterator
-						.next();
+			for ( Iterator<AbstractMap.SimpleEntry<String,String>> iterator = params.iterator(); iterator.hasNext(); ) {
+				AbstractMap.SimpleEntry<String,String> entry = iterator.next();
 				try {
 					if(StringUtils.isNotBlank(entry.getValue())) {
 						url.append(entry.getKey())
--- a/src/main/java/com/openshift/internal/restclient/api/capabilities/PodExec.java
+++ b/src/main/java/com/openshift/internal/restclient/api/capabilities/PodExec.java
@@ -1,0 +1,177 @@
+/*******************************************************************************
+ * Copyright (c) 2015 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package com.openshift.internal.restclient.api.capabilities;
+
+import com.openshift.internal.restclient.DefaultClient;
+import com.openshift.internal.restclient.URLBuilder;
+import com.openshift.internal.restclient.capability.AbstractCapability;
+import com.openshift.internal.restclient.okhttp.ResponseCodeInterceptor;
+import com.openshift.internal.restclient.okhttp.WebSocketAdapter;
+import com.openshift.restclient.IApiTypeMapper;
+import com.openshift.restclient.IClient;
+import com.openshift.restclient.capability.IStoppable;
+import com.openshift.restclient.api.capabilities.IPodExec;
+import com.openshift.restclient.http.IHttpConstants;
+import com.openshift.restclient.model.IPod;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+import okhttp3.ResponseBody;
+import okhttp3.ws.WebSocket;
+import okhttp3.ws.WebSocketCall;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class PodExec extends AbstractCapability implements IPodExec {
+
+	private static final Logger LOG = LoggerFactory.getLogger(IPodExec.class);
+	private static final String CAPABILITY = "exec";
+
+	private static final String COMMAND = "command";
+
+	private static final String K8S_PROTOCOL_HEADER = "X-Stream-Protocol-Version";
+
+	private static final String K8S_PROTOCOL = "channel.k8s.io";
+
+	public static final int CHANNEL_STDOUT = 1;
+	public static final int CHANNEL_STDERR = 2;
+	public static final int CHANNEL_EXECERR = 3;
+
+	private final IPod pod;
+	private final DefaultClient client;
+	private final IApiTypeMapper mapper;
+
+	public PodExec(IPod pod, IClient client) {
+		super( pod, client, CAPABILITY );
+		this.pod = pod;
+		this.client = client.adapt(DefaultClient.class);
+		this.mapper = client.adapt(IApiTypeMapper.class);
+	}
+	
+	@Override
+	public String getName() {
+		return PodExec.class.getSimpleName();
+	}
+
+	@Override
+	public IStoppable start( IPodExecOutputListener listener, Options options, String... commands ) {
+
+		if ( options == null ) {
+			options = new Options();
+		}
+
+		Map<String,String> parameters = options.getMap();
+
+		OkHttpClient okClient = client.adapt(OkHttpClient.class);
+
+		URLBuilder urlBuilder = new URLBuilder(client.getBaseURL(), mapper)
+				.resource(pod)
+				.subresource(CAPABILITY)
+				.addParameters(parameters);
+
+		// The main command and all arguments are specified as 'command' parameters
+		for ( String command : commands ) {
+			urlBuilder.addParmeter( COMMAND, command );
+		}
+
+		final String endpoint = urlBuilder.websocket();
+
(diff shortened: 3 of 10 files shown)


src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -243,18 +243,23 @@ public String getServerReadyStatus() {
 	}
 
 	public Request.Builder newRequestBuilderTo(String endpoint){
+		return newRequestBuilderTo(endpoint, MEDIATYPE_APPLICATION_JSON);
+	}
+
+	public Request.Builder newRequestBuilderTo(String endpoint,String acceptMediaType){
```
jupierce: The exec endpoint does not return JSON, so I needed to permit new accept types.

src/main/java/com/openshift/internal/restclient/URLBuilder.java
```diff
@@ -42,7 +41,7 @@
 	private String baseUrl;
 	private String kind;
 	private String name;
-	private Map<String, String> params = new HashMap<String, String>();
+	private ArrayList<AbstractMap.SimpleEntry<String,String>> params = new ArrayList<>();
 	private final IApiTypeMapper typeMappings;
```
jupierce: The exec endpoint takes multiple "command" parameters. The first specifies the binary to run. Subsequent command parameters specify arguments. Changes enable this pattern in URLBuilder.

src/main/java/com/openshift/internal/restclient/okhttp/ResponseCodeInterceptor.java
```diff
@@ -56,7 +61,9 @@ public Response intercept(Chain chain) throws IOException {
 				response = makeSuccessIfAuthorized(response);
 				break;
 			default:
-				throw createOpenShiftException(client, response, null);
+				if ( response.request().tag() instanceof Ignore == false ) {
```
jupierce: For websocket calls, if the interceptor throws an exception, it is swallowed up in the OkHttp Dispatcher. It is never delivered to WebSocketListener.onFailure. In order for ExecOutputListenerAdapter to indicate failure reliably, I needed a way to have the interceptor ignore my requests.

jcantrill:
Let me try to rebase and merge this https://github.com/openshift/openshift-restclient-java/pull/205 which sort of fits into these additional endpoints.

jcantrill:
[test]  I suspect you will need to rebase.  Also, please consider writing an integration test to prove out the functionality.  There are several examples in the test branch which I have found useful when unit tests just don't seem to cover everything

src/main/java/com/openshift/internal/restclient/URLBuilder.java
jcantrill: Is this because you can specify certain parameters multiple times which a Map obviously would not support?

src/main/java/com/openshift/internal/restclient/capability/resources/PodExec.java
```diff
@@ -0,0 +1,156 @@
...
+import okhttp3.ws.WebSocket;
+import okhttp3.ws.WebSocketCall;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class PodExec implements IPodExec {
```
jcantrill: Consider extending from https://github.com/openshift/openshift-restclient-java/blob/master/src/main/java/com/openshift/internal/restclient/capability/AbstractCapability.java which will ensure the apigroups are checked for supportability

src/main/java/com/openshift/internal/restclient/capability/resources/PodExec.java
```diff
@@ -0,0 +1,156 @@
...
+ * Copyright (c) 2015 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package com.openshift.internal.restclient.capability.resources;
```
jcantrill: Please repackage here: com.openshift.internal.restclient.api.capabilities  as I'm trying to align more with how resources are discovered from the server.

src/main/java/com/openshift/restclient/capability/resources/IPodExec.java
```diff
@@ -0,0 +1,150 @@
...
+ * Copyright (c) 2016 Red Hat, Inc. 
+ * Distributed under license by Red Hat, Inc. All rights reserved. 
+ * This program is made available under the terms of the 
+ * Eclipse Public License v1.0 which accompanies this distribution, 
+ * and is available at http://www.eclipse.org/legal/epl-v10.html 
+ * 
+ * Contributors: 
+ * Red Hat, Inc. - initial API and implementation 
+ ******************************************************************************/
+package com.openshift.restclient.capability.resources;
```
jcantrill: please repackage here: com.openshift.restclient.api.capabilities to align more with where resources are on the server

src/main/java/com/openshift/restclient/capability/resources/IPodExec.java
```diff
@@ -0,0 +1,150 @@
...
+		 * Exec received stderr message
+		 * @param message
+		 */
+		void onStdErr(String message);
+
+		/**
+		 * Exec (channel 3) error message
+		 * @param message
+		 */
+		void onExecErr(String message);
```
jcantrill: Will you actually get a call here instead of just getting 'onStdErr'?

src/main/java/com/openshift/internal/restclient/capability/resources/PodExec.java
```diff
@@ -0,0 +1,156 @@
...
+
+	@Override
+	public IStoppable start( IPodExecOutputListener listener, Options options, String... commands ) {
+
+		Map<String,String> parameters = options.getMap();
+
+		OkHttpClient okClient = client.adapt(OkHttpClient.class);
+
+		URLBuilder urlBuilder = new URLBuilder(client.getBaseURL(), mapper)
+				.kind(pod.getKind())
```
jcantrill: Your could replace these with builder.resource(pod)

src/main/java/com/openshift/internal/restclient/URLBuilder.java
```diff
@@ -105,7 +104,7 @@ public URLBuilder resource(IResource resource) {
 	}
 
 	public URLBuilder addParmeter(String key, String value) {
-		params.put(key, value);
+		params.add(new AbstractMap.SimpleEntry<>( key, value ));
```
jcantrill: Is there a URLBuilderTest to support this change?

src/main/java/com/openshift/internal/restclient/capability/resources/PodExec.java
```diff
@@ -0,0 +1,156 @@
...
+					listener.onStdOut(msg);
+					break;
+				case 2:
+					listener.onStdErr(msg);
+					break;
+				case 3:
+					listener.onExecErr(msg);
+					break;
+				default:
+					LOG.warn("Unable to deliver exec message of type [" + type + "]: " + msg );
```
jcantrill: warn will actually take a sprintf type format: http://www.slf4j.org/apidocs/org/slf4j/Logger.html

src/main/java/com/openshift/internal/restclient/okhttp/ResponseCodeInterceptor.java
jcantrill: Hmm... A not so obvious way to pass around a context.  Okhttp api docs could be better... WFM

jcantrill:
After reviewing the code there should be minimal impact from the PR I referenced.

src/main/java/com/openshift/internal/restclient/URLBuilder.java
jupierce: Haven't done any test work yet. Wanted to ensure you were satisfied with the direction.

src/main/java/com/openshift/internal/restclient/URLBuilder.java
jupierce: Yep:
$ cat /tmp/file
would translate into the URL:
....exec?command=cat&command=/tmp/file....

src/main/java/com/openshift/internal/restclient/capability/resources/PodExec.java
jupierce: Included in next push.

src/main/java/com/openshift/internal/restclient/capability/resources/PodExec.java
jupierce: Next push.

src/main/java/com/openshift/restclient/capability/resources/IPodExec.java
jupierce: Next push.

src/main/java/com/openshift/restclient/capability/resources/IPodExec.java
jupierce: I can definitely get talk on that channel. Example of launching non-existent binary:
```
OPEN!
stdout: 
stderr: exec: "/bin/doesnotexist": stat /bin/doesnotexist: no such file or directory

execerr: error executing command in container: Error executing in Docker Container: -1
CLOSE!
```

src/main/java/com/openshift/internal/restclient/capability/resources/PodExec.java
jupierce: Nice. Next push.

src/main/java/com/openshift/internal/restclient/capability/resources/PodExec.java
jupierce: Next push.

src/main/java/com/openshift/internal/restclient/URLBuilder.java
jcantrill: wfm as long as its backwards compatible. I am unable to think of any place this might be an issue...but..just in case

jupierce: Next push.

src/main/java/com/openshift/restclient/api/capabilities/IPodExec.java
```diff
@@ -0,0 +1,150 @@
...
+public interface IPodExec extends ICapability{
+
+	/**
+	 * Execute a command on a named container in this pod
+	 * @param listener Listener for command output
+	 * @param options Options for the exec
+	 * @param commands A command to run and any arguments
+	 * @return A Handle to allow termination of the connection
+	 */
+	IStoppable start( IPodExecOutputListener listener, Options options, String... commands );
```
gabemontero: There's a implied understanding that the first item in the commands string is the command ... should we be more explicit and have a separate parameter for the command, and then a `String...` for the args

src/main/java/com/openshift/internal/restclient/api/capabilities/PodExec.java
```diff
@@ -0,0 +1,151 @@
...
+		@Override
+		public void onFailure(IOException e, Response response) {
+			listener.onFailure(e);
+		}
+
+		@Override
+		public void onMessage(ResponseBody message) throws IOException {
+			int type = message.byteStream().read();
+			String msg = message.string();
+			switch ( type ) {
```
gabemontero: What dictates that type 1, 2, or 3 map to stdout, stderr, execerr ?  If in this pull I'm missing it / please advise.  Otherwise, a comment stating what controls that might be useful for other folks who view this in the future.

jcantrill: It may actually be more useful to extract them to constants for more meaning

jupierce: Constants extracted. 
@gabemontero The protocol is proprietary to k8s. Added a comment to onMessage.

jupierce:
@jcantrill Working on integration test now.

src/main/java/com/openshift/restclient/api/capabilities/IPodExec.java
jupierce: I'm all for explicit, but I'd argue the more command idiom here is a single array: 
https://docs.oracle.com/javase/7/docs/api/java/lang/ProcessBuilder.html#command(java.lang.String...)

https://docs.oracle.com/javase/7/docs/api/java/lang/Runtime.html#exec(java.lang.String[])

gabemontero: Fair point - there is precedence.  The fact that I don't agree with it is another thing ;-) but sure, stay consistent with JSE.

src/main/java/com/openshift/internal/restclient/api/capabilities/PodExec.java
```diff
@@ -0,0 +1,177 @@
...
+		final String endpoint = urlBuilder.websocket();
+
+		Request request = client.newRequestBuilderTo(endpoint, IHttpConstants.MEDIATYPE_ANY)
+				.method( "GET", null )
+				.addHeader(K8S_PROTOCOL_HEADER, K8S_PROTOCOL )
+				// Unless we mark this as ignored, exceptions triggered by interceptor would be lost in dispatcher thread
+				.tag( new ResponseCodeInterceptor.Ignore() {} )
+				.build();
+
+		WebSocketCall call = WebSocketCall.create(okClient, request);
```
gabemontero: just a tangential observation for @jupierce - the websocket stuff is what the pod log retrieval stuff uses ...the jenkins-plugin started leveraging pod log retrieval from the restclient a month or 2 ago.

jupierce:
@jcantrill PodExecIntegrationTest pushed. I've found the k8s doesn't respect stdout/stderr flags, so I'm not testing for those at the moment.

src/test/java/com/openshift/internal/restclient/api/capabilities/PodExecIntegrationTest.java
```diff
@@ -0,0 +1,207 @@
...
+		public final List<String> messages = Collections.synchronizedList( new ArrayList<String>() );
+
+		@Override
+		public void onOpen() {
+			assertTrue(openCalled.compareAndSet(false,true));
+		}
+
+		@Override
+		public void onStdOut(String message) {
+			System.err.println( "onStdOut: " + message );
```
jcantrill: Please replace with logger... debug level?  maybe?

src/test/java/com/openshift/internal/restclient/api/capabilities/PodExecIntegrationTest.java
```diff
@@ -0,0 +1,207 @@
...
+		@Override
+		public void onClose(int code, String reason) {
+			assertTrue(closeCalled.compareAndSet(false,true));
+			testDone.countDown();
+		}
+
+		@Override
+		public void onFailure(IOException e) {
+			failureCalled.set(true);
+			e.printStackTrace();
```
jcantrill: logger.error

src/test/java/com/openshift/internal/restclient/api/capabilities/PodExecIntegrationTest.java
```diff
@@ -0,0 +1,207 @@
...
+			testDone.countDown();
+		}
+
+	}
+
+	@Before
+	public void setUp() throws Exception {
+		System.setProperty(IBinaryCapability.OPENSHIFT_BINARY_LOCATION, helper.getOpenShiftLocation());
```
jcantrill: You shouldnt need this since your impl does not depend on oc

jcantrill:
aside from the few test nits...LGTM

src/test/java/com/openshift/internal/restclient/api/capabilities/PodExecIntegrationTest.java
jupierce: Removed in next push.

src/test/java/com/openshift/internal/restclient/api/capabilities/PodExecIntegrationTest.java
jupierce: Next push.

src/test/java/com/openshift/internal/restclient/api/capabilities/PodExecIntegrationTest.java
jupierce: WFM. Next push.

jupierce:
ptal @jcantrill

jcantrill:
LGTM  [merge]

PR #221, opened 2016-09-30
add setter for proxy authenticator

jcantrill:
@adietish @fbricon @jeffmaury  Is this any better of an implementation to satisfy the issues I raised in #199 ?

### DIFF
--- a/src/main/java/com/openshift/restclient/ClientBuilder.java
+++ b/src/main/java/com/openshift/restclient/ClientBuilder.java
@@ -33,6 +33,7 @@ import com.openshift.internal.restclient.okhttp.ResponseCodeInterceptor;
 import com.openshift.restclient.http.IHttpConstants;
 import com.openshift.restclient.utils.SSLUtils;
 
+import okhttp3.Authenticator;
 import okhttp3.Dispatcher;
 import okhttp3.OkHttpClient;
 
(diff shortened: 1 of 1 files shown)


jcantrill:
[test]

jeffmaury:
I don't much added value because although from the API point of view, no okhttp classes are exposed, when it comes to implementation, then you need to implement an okhttp interface

jcantrill:
This is true but we are leaving it to whomever wishes to set the proxy and they will be externally dependent on the authenticator.  I'm conflicted a little, but looking at #199 it is apparent that we already expose those libs in ClientBuilder so maybe it REALLY is immaterial we are doing it here.  Honestly, I don't see us moving to yet a third client.  At best, we might hide the fabric8 client which also depends on okhttp.  Any thoughts on closing this PR and just accepting #199 ?

adietish:
When I started looking into details I saw OpenShiftAuthenticator which implements (okHTTP) Authenticator which is an exposed and hardwired dependency on okHttp. I think that we should either not expose any okHTTP classes or just dont care. Unless this PR is a step in decoupling?

jcantrill:
@adietish This was kind of my point.  This PR is intended to be a step at decoupling where a user like in #199 can provide their own proxy authenticator, which in this case, they would need to know what the underlying dependency is.  We, however, expose the client (sort of) in the imports of ClientBuilder. I am unable to conceive of a reason we would switch to yet a 3rd client which leaves me in a position of ambivalence; I'm not certain I care enough about the exposure.  That makes me ask if we should just accept #199 and skip this PR.  Is it really worth the decoupling as I have presented it here?  I'm not hugely inclined to generate a bunch of interface classes to hide okhttp

PR #222, opened 2016-09-30
fix list returning objects.  It's not a thing

jcantrill:
@jeffmaury @adietish  fix to list that was talked about earlier.  I'm not sure why 'object' was ever in there as I can not find reference to it in the origin code base.


jeffmaury:
Objects are used in templates, no? So maybe not in that case but you should find it in Origin code

jcantrill:
@jeffmaury objects are used for templates. but this change is for List.  Ran the server template integration test which passes without failure.

adietish:
+1 works for me

jcantrill:
[merge]

PR #223, opened 2016-10-04
Fixes for container command args and service port names

transamericamoon:
Add "command" and "args" attribute to IContainer, and add ability to supply port name to the IService.addPort method.

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/model/Container.java
+++ b/src/main/java/com/openshift/internal/restclient/model/Container.java
@@ -15,18 +15,20 @@ import static com.openshift.internal.util.JBossDmrExtentions.*;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 
-import com.openshift.restclient.model.ILifecycle;
 import org.jboss.dmr.ModelNode;
+import org.jboss.dmr.ModelType;
 
 import com.openshift.internal.restclient.model.properties.ResourcePropertyKeys;
 import com.openshift.internal.restclient.model.volume.EmptyDirVolume;
 import com.openshift.internal.restclient.model.volume.VolumeMount;
 import com.openshift.restclient.images.DockerImageURI;
 import com.openshift.restclient.model.IContainer;
+import com.openshift.restclient.model.ILifecycle;
 import com.openshift.restclient.model.IPort;
 import com.openshift.restclient.model.volume.IVolume;
 import com.openshift.restclient.model.volume.IVolumeMount;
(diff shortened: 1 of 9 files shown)


src/main/java/com/openshift/restclient/model/IContainer.java
```diff
@@ -56,6 +58,12 @@
 	void setLifecycle(ILifecycle lifecycle);
 	ILifecycle getLifecycle();
 	
+    void setCommand(Set<String> command);
+    Set<String> getCommand();
+    
+    void setCommandArgs(Set<String> args);
```
jcantrill: Might this better be implemented as a String... or a list since Set does not guarantee order.  Is order relevent?

src/main/java/com/openshift/restclient/model/IContainer.java
```diff
@@ -56,6 +58,12 @@
 	void setLifecycle(ILifecycle lifecycle);
 	ILifecycle getLifecycle();
 	
+    void setCommand(Set<String> command);
```
jcantrill: Might this better be implemented as a String... or a list since Set does not guarantee order. Is order relevent?

src/main/java/com/openshift/restclient/model/IService.java
```diff
@@ -40,6 +40,8 @@
 	
 	IServicePort addPort(int port, int targetPort);
 	
+	IServicePort addPort(int port, int targetPort, String name);
```
jcantrill: Please add a supporting test

transamericamoon:
Ok, converted to List and add Tests.

src/main/java/com/openshift/internal/restclient/model/Container.java
```diff
@@ -136,11 +139,31 @@ public void setPorts(Set<IPort> ports) {
...
 	}
-
+	
 	@Override
 	public String getImagePullPolicy() {
 		return asString(node, propertyKeys, IMAGE_PULL_POLICY);
 	}
+	
+	@Override
+    public void setCommand(Set<String> command) {
```
transamericamoon: from what i understand, command is equivalent to DockerFile EntryPoint and args is equivalent to dockerfile CMD.  Both are list.

src/main/java/com/openshift/internal/util/JBossDmrExtentions.java
```diff
@@ -177,6 +177,43 @@ public static Set asSet(ModelNode root, Map<String, String []> propertyKeys, Str
...
+     * such as command Args for containers.  
+     *  
+     * @param root
+     * @param propertyKeys
+     * @param key
+     * @param type
+     * @return
+     */
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    public static List asList(ModelNode root, Map<String, String []> propertyKeys, String key, ModelType type){
```
jcantrill: Please add supporting test

transamericamoon: Test added.

jcantrill:
LGTM [merge]

PR #224, opened 2016-10-05
Expose errors during async log retrieval

jupierce:
Exceptions when attempting to retrieve logs asynchronously from pods are currently swallowed in the okhttp dispatcher thread. Perhaps we could introduce a model similar to #220  ?

fyi @gabemontero


jupierce:
To enable fix for openshift/jenkins-plugin#74

jcantrill:
LGTM [merge]

PR #227, opened 2016-10-07
add service account to replication controller

transamericamoon:
add service account to replication controller

Ref: https://docs.openshift.com/enterprise/3.2/dev_guide/deployments.html#run-pod-with-different-service-account

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/model/ReplicationController.java
+++ b/src/main/java/com/openshift/internal/restclient/model/ReplicationController.java
@@ -51,6 +51,7 @@ public class ReplicationController extends KubernetesResource implements IReplic
 	protected static final String SPEC_REPLICAS = "spec.replicas";
 	protected static final String SPEC_SELECTOR = "spec.selector";
 	protected static final String STATUS_REPLICA = "status.replicas";
+	protected static final String SERVICEACCOUNTNAME = "spec.template.spec.serviceAccountName";
 
 	protected static final String IMAGE = "image";
 	protected static final String ENV = "env";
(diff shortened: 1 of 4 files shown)


src/main/java/com/openshift/internal/restclient/model/ReplicationController.java
```diff
@@ -364,4 +366,24 @@ public void addVolume(IVolumeSource volumeSource) {
 		}
 		volList.add(ModelNode.fromJSONString(volumeSource.toJSONString()));
 	}
+
+    @Override
+    public void setServiceAccount(String serviceAccount) {
```
jcantrill: Why do we have these getters and setters?  I dont see that as an available part of the pod spec

jcantrill:
[test]

src/main/java/com/openshift/internal/restclient/model/ReplicationController.java
transamericamoon: Are you saying we need to add it into the IPod interface/Impl as well? I was kind of wondering why the RC does not implement the pod interface.

jcantrill: RC does not implement Pod because it is not a Pod; we need to keep some relative relationship between the client and server code.  What I am saying is for RC there is no entry in the json tree for 'serviceaccount'.  There is no need for this method to exist as far as i can see as it will not provide any useful information to the controllers on the server: https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/pkg/api/types.go#L1576

transamericamoon: Thanks for the info, I was confused because the openshift docs show both "serviceAccount" and "serviceAccontName" should be fixed now with get/set test.

jcantrill:
[merge]

PR #231, opened 2016-10-16
Added setter and getter for Service type

l0rd:
Added getter and setter for [Service type](http://kubernetes.io/docs/user-guide/services/#publishing-services---service-types)


jcantrill:
[merge]

PR #233, opened 2016-10-19
add configmap back in

### LINKED ISSUE #208: ResourceFactory unable to create ConfigMap instances
I'm having an issue in my client using a watcher to watch config maps after upgrading to 5.1.0.Final. My watcher listener code looks like the following:

```java
    public void connected(List<IResource> resources) {
        for (IResource resource : resources) {
       
            IConfigMap configMap = (IConfigMap) resource;
        }
    }
```

The cast worked earlier, but failed on 5.1.0.Final. After 4.0.5, the ConfigMap kind was explicitly removed from ResourceFactory (https://github.com/openshift/openshift-restclient-java/commit/557d5ee492ec4d6bafb9d1fa5006afcab8e421b2), and I suppose it was to be created using this new typeMapper. However, after reading the code of the WatcherClient and ResourceFactory, I can't really see that the typemapper is used when receiving the response. The ResourceFactory seems to still use the IMPL_MAP mapping where the configmap kind does not exist.

jcantrill:
fixes #208 

cc @lulf 

[test]

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/ResourceFactory.java
+++ b/src/main/java/com/openshift/internal/restclient/ResourceFactory.java
@@ -25,6 +25,7 @@ import org.jboss.dmr.ModelNode;
 
 import com.openshift.internal.restclient.model.Build;
 import com.openshift.internal.restclient.model.BuildConfig;
+import com.openshift.internal.restclient.model.ConfigMap;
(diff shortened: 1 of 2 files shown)


jcantrill:
[merge]

PR #234, opened 2016-10-27
jbide-23184: Add Capability to Watch Projects

bdshadow:


jcantrill:
[test]  It's been awhile since I was in this part of the code but I would have expected a larger change.

adietish:
+1, looks good.
This should fix the endless looping we currently have if watching projects isnt supported.
I specifically like the comment pointing on why we swallow and the unit test that asserts this behaviour.

adietish:
[merge]

PR #236, opened 2016-10-28
Fixed NPE

luksa:
WatchClient.watch() passes null as the response, which resulted in an NPE


jcantrill:
[merge]

adietish:
@luksa great that you found this one. Thanks for fixing it.
It would have been great if your commit message was more telling, that would allow to know where ithe NPE happens without inspecting the github PR or actual patch.

PR #237, opened 2016-11-02
jbide-23358: OpenShift Explorer: Pod status is not reflecting reality

bdshadow:


src/main/java/com/openshift/internal/restclient/model/Pod.java
```diff
@@ -61,10 +74,38 @@ public String getHost() {
...
+	
+	/**
+	 * The logic of the method is a copied from 'podStatus' function of
+	 * [app/scripts/filters/resources.js] of [openshift/origin-web-console]
+	 */
 
 	@Override
 	public String getStatus() {
-		return asString(POD_STATUS);
+		if (get(POD_DELETION_TIMESTAMP).isDefined()) {
```
jcantrill: Supporting test?

jcantrill: I thought it existed in JbossDmrExtentions, but it is not, would be to use has(NODE_PATH) because get coupled with 'isdefined' actually creates the child node where 'has' only checks for it.  Depending on the urgency of the PR, I would suggest adding 'has(String...)' to the utility class

src/main/java/com/openshift/internal/restclient/model/Pod.java
```diff
@@ -61,10 +74,38 @@ public String getHost() {
...
+						return status;
+					}
+				}
+			}
+		}
+		return get(POD_STATUS_REASON).isDefined() ? asString(POD_STATUS_REASON) : asString(POD_STATUS_PHASE);
+	}
+	
+	private String getContainerStatusIfDefined(ModelNode containerStatus, String path, String statusPrefix) {
+		ModelNode node = containerStatus.get(getPath(path));
```
jcantrill: JbossDmrExtentions.asString(Node,Map, path) in lieu of this

src/main/java/com/openshift/internal/restclient/model/Pod.java
bdshadow: @jcantrill 
Yes, i thought about the tests. 
The existing `PodTest` is actually going through all the loops over `containerStatuses`.
In order to test it properly we need to create more `v1_pod` jsons for every situation. I'm ok with it, but maybe you can offer a more nice option

jcantrill: There is a class called ModelNodeBuilder where you could build and manipulate the values you desire and instantiate a Pod directly.  This would allow you to make changes without needing to craft permutations of the JSON string

bdshadow: @jcantrill i don't think we need to add it to `JbossDmrExtentions`, because `KubernetesResource` has `has(String key)` method itself, and we can check right from resources.
Anyway i'll change my `isDefined` to this `has`, as it really creates the child node. Didn't know about it

src/main/java/com/openshift/internal/restclient/model/Pod.java
```diff
@@ -61,10 +76,35 @@ public String getHost() {
...
+	
+	/**
+	 * The logic of the method is a copied from 'podStatus' function of
+	 * [app/scripts/filters/resources.js] of [openshift/origin-web-console]
+	 */
 
 	@Override
 	public String getStatus() {
-		return asString(POD_STATUS);
+		if (has(POD_DELETION_TIMESTAMP)) {
```
adietish: not complaining about your code here: to me all this feels like the backend is inconsistent and should get fixed. I believe that the (server reported) json is missing an (additional) unique field where the pod state is reflected. The whole method inspects different locations and merge these to exactly this: a single and unique field for the pod state. Shouldn't we file an issue to openshift/origin?

src/main/java/com/openshift/internal/restclient/model/Pod.java
```diff
@@ -61,10 +76,35 @@ public String getHost() {
...
+			return "Terminating";
```
adietish: I also see us adding a state here, that doesnt exist in openshift/origin. Shouldnt we file this issue?

src/main/java/com/openshift/internal/restclient/model/Pod.java
```diff
@@ -61,10 +76,35 @@ public String getHost() {
...
+		}
+		ModelNode node = get(POD_STATUS_CONTAINER_STATUSES);
+		if (node.getType() == ModelType.LIST) {
+			for (ModelNode containerStatus : node.asList()) {
+				for (String containerReason: POD_STATUS_CONTAINER_STATES.keySet()) {
```
adietish: this (l93 - l98) could be extracted to a method #getPodContainerState for improved readability

adietish:
[test]

src/main/java/com/openshift/internal/restclient/model/Pod.java
bdshadow: @adietish i agree with you. It's really strange, that they do all these logic in web-console webpage themselves.
However, if we create an issue and ask to add this unique field, they can say that this field is a duplicate of already existing ones

src/main/java/com/openshift/internal/restclient/model/Pod.java
bdshadow: @adietish [it exists](https://github.com/openshift/origin-web-console/blob/7c57218bb83964ba0aab86ebf6ee42fb52aabf39/app/scripts/filters/resources.js#L1210). But again - not in the os-origin, but in os-web-console.

adietish: @bdshadow that's exactly my point. I think that this should be in origin, not in tools (Eclipse or Web-UI). This makes me feel like they're lacking things if user code(s) have to add this state and  have extensive logic to determine a pod state (that's then replicated across these tools)

src/main/java/com/openshift/internal/restclient/model/Pod.java
adietish: @bdshadow yes but then the extensive logic that gets replicated across tools shows to me, that this belongs into the server, not into tools. Duplication will happen, but better have them in one single place than several times in all tools.

adietish:
[test]

adietish:
[merge]

PR #238, opened 2016-11-08
JBIDE-23463 Properties: Build pod has state of exit code 0 instead of correct status completed

scabanovich:
Logic of Pod.getStatus() is fixed to copy that of
resources.js .filter('podStatus')

### DIFF
--- a/src/test/java/com/openshift/internal/restclient/model/v1/PodTest.java
+++ b/src/test/java/com/openshift/internal/restclient/model/v1/PodTest.java
@@ -81,6 +81,19 @@ public class PodTest {
 		assertEquals(pod.getStatus(), "ReasonToTerminate");
 	}
 	
+	/**
+	 * Check that if both reason and exit code are set, status returns the reason.
+	 */
+	@Test
+	public void testGetStatusTerminateReasonAndExit() {
+		ModelNode node = new ModelNode();
+		node.get("reason").set("ReasonToTerminate");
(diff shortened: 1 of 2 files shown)


jcantrill:
Looks good. Are there tests to be added as well? [test]

adietish:
@scabanovich thx for fixing this!
You might want to correct the commit msg: JBIDE-21463 -> JBIDE-23463
And yes, it would be even greater if you could add tests. PodTest currently passes while it shouldnt, since we consider the current behavior as faulty. So there's at least a missing assertion for the current faulty behavior

scabanovich:
I have added some tests.

src/test/java/com/openshift/internal/restclient/model/v1/PodTest.java
```diff
@@ -81,13 +81,39 @@ public void testGetStatusTerminateReason() {
 		assertEquals(pod.getStatus(), "ReasonToTerminate");
 	}
 	
+	/**
+	 * Check that if both reason and exit code are set, status returns the reason.
+	 */
+	@Test
+	public void testGetStatusTerminateReasonAndExit() {
+		ModelNode node = new ModelNode();
```
jcantrill: LGTM. In future, there is a ModelNodeBuilder class that might make it easier for your to build up the json.

jcantrill:
[merge]

PR #239, opened 2016-11-09
[OSJC-274] - When creating an app from a builder image template, expanding the buildconfig causes

jeffmaury:


jcantrill:
[merge]

PR #247, opened 2016-12-15
client side filtering of get requests

### LINKED ISSUE #246: Serverside filtering of GET requests
When doing a GET request with labels the filtering is now done clientside. There is a todo in the code to figure out how to do this serverside. 

All GET queries to the oapi endpoints support a labelQuery queryparameter. So in my eyes a solution could be to convert labels to csv string and add them as a query paramter to the URL builder. 

I have done an attempt in my fork on how to do this. I have not managed to get the tests to run quite yet so there might be some small bugs. 

It should be noted that labelQuery supports features that cannot be directly mapped to a map of key=value. I.E this is a valid labelQuery "label2,!label3". This basically says fetch resources that have label2 set but not label3. So it would perhaps be wise to add a method to get kinds with a predefined labelQuery as a string.

bjartek:
Trying to fix #246

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/DefaultClient.java
+++ b/src/main/java/com/openshift/internal/restclient/DefaultClient.java
@@ -8,6 +8,8 @@
  ******************************************************************************/
 package com.openshift.internal.restclient;
 
+import static java.util.stream.Collectors.joining;
+
 import static com.openshift.internal.restclient.capability.CapabilityInitializer.initializeClientCapabilities;
 
 import java.io.IOException;
@@ -15,6 +17,7 @@ import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -123,21 +126,31 @@ public class DefaultClient implements IClient, IHttpConstants{
 	@SuppressWarnings("unchecked")
 	@Override
 	public <T extends IResource> List<T> list(String kind, String namespace, Map<String, String> labels) {
-		IList list = get(kind, namespace);
-		ArrayList<T> items = new ArrayList<T>((Collection<T>)list.getItems());
-		return filterItems(items, labels); //client filter until we can figure out how to restrict with a server call
-	}
-	
-	private <T extends IResource> List<T> filterItems(List<T> items, Map<String, String> labels){
-		if(labels.isEmpty()) return items;
-		List<T> filtered = new ArrayList<T>();
-		for (T item : items) {
-			if( item.getLabels().entrySet().containsAll(labels.entrySet())){
-				filtered.add(item);
-			}
+
+		String labelQuery="";
+		if(labels != null && !labels.isEmpty()){
+			 labelQuery = labels.entrySet().stream()
+					.map(e -> e.getKey() + "=" + e.getValue())
+					.collect(joining(","));
 		}
-		return filtered;
+		return list(kind, namespace, labelQuery);
 	}
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public <T extends IResource> List<T> list(String kind, String namespace, String labelQuery) {
+
+		Map<String, String> params = new HashMap<>();
+		if(labelQuery != null && !labelQuery.isEmpty()){
+			params.put("labelSelector", labelQuery);
+		}
+
+		IList resources = execute(HttpMethod.GET.toString(), kind, namespace, null, null, null, params);
+		List<T> items = new ArrayList<>();
+		items.addAll((Collection<? extends T>) resources.getItems());
+		return items;
+	}
+
 	
 	@Override
 	public Collection<IResource> create(IList list, String namespace){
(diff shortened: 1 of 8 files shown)


bjartek:
Ups. Looks like my formatting is off. What codestyle do you use?

jcantrill:
@bjartek I don't have any defined code style other then the default 'java conventions' from Eclipse.

bjartek:
Some of the tests are not valid anymore since the filtering is done in the api now. Stil needs to format things better it looks like.

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -123,21 +126,21 @@ public String getResourceURI(IResource resource) {
...
-	private <T extends IResource> List<T> filterItems(List<T> items, Map<String, String> labels){
-		if(labels.isEmpty()) return items;
-		List<T> filtered = new ArrayList<T>();
-		for (T item : items) {
-			if( item.getLabels().entrySet().containsAll(labels.entrySet())){
-				filtered.add(item);
-			}
+
+		Map<String, String> params = new HashMap<>();
+		if(!labels.isEmpty()){
```
jcantrill: Its worth adding a null check here too.

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -123,21 +126,31 @@ public String getResourceURI(IResource resource) {
 	@SuppressWarnings("unchecked")
 	@Override
 	public <T extends IResource> List<T> list(String kind, String namespace, Map<String, String> labels) {
-		IList list = get(kind, namespace);
-		ArrayList<T> items = new ArrayList<T>((Collection<T>)list.getItems());
-		return filterItems(items, labels); //client filter until we can figure out how to restrict with a server call
+
+		String labelQuery="";
+		if(!labels.isEmpty()){
```
jcantrill: Null check

bjartek:
I added another method to list to take a predefined labelQuery so that it is possible to do lists that check if a label is present or not present.

bjartek:
I will add some teste in a little bit, just have to do some meetings at work first :)

bjartek:
Are these integration tests enough?

src/test/java/com/openshift/internal/restclient/DefaultClientFilterIntegrationTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+		assertEquals(1, list.size());
+		IBuildConfig bc = list.get(0);
+		assertEquals("build1", bc.getName());
+	}
+
+
+
+	private BuildConfig createBuildConfigWithLabels(IProject project, String name, HashMap<String, String> labelFilter) {
+		BuildConfig bc = factory.stub(BUILD_CONFIG, name);
```
jcantrill: Consider replacing with the the capability builder as described here: https://github.com/openshift/openshift-restclient-java

src/test/java/com/openshift/internal/restclient/DefaultClientFilterIntegrationTest.java
```diff
@@ -0,0 +1,155 @@
...
+		if (labelFilter != null && !labelFilter.isEmpty()) {
+			for (Map.Entry<String, String> label : labelFilter.entrySet()) {
+				bc.addLabel(label.getKey(), label.getValue());
+			}
+		}
+		bc = client.create(bc, project.getNamespace());
+		assertNotNull("Exp. the bc to be found but was not",
+				waitForResource(client, BUILD_CONFIG, project.getName(), bc.getName(),
```
jcantrill: Is this necessary?  I would have expected the create call to be sufficent

src/test/java/com/openshift/internal/restclient/DefaultClientFilterIntegrationTest.java
```diff
@@ -0,0 +1,155 @@
...
+			put("bar", "yes");
+		}}));
+
+		buildConfigs.add(createBuildConfigWithLabels(project, "build4", new HashMap<>()));
+
+	}
+
+	@AfterClass
+	public void cleanup() {
+		buildConfigs.forEach(bc -> cleanUpResource(client, bc));
```
jcantrill: You should be able to just delete the project if you dont want to make individual calls.

src/test/java/com/openshift/internal/restclient/DefaultClientTest.java
```diff
@@ -97,6 +98,8 @@ private DefaultClient givenClient(URL baseUrl, String token, String user) {
 	
 	@SuppressWarnings("serial")
 	@Test
+
+	@Ignore //filtering is done serverside
```
jcantrill: I would remove these tests all together since we have an integration test that supports.  Integration will be enabled on a regular basis hopefully after the first of the year

src/test/java/com/openshift/internal/restclient/DefaultClientFilterIntegrationTest.java
```diff
@@ -0,0 +1,155 @@
...
+	private IProject project;
+
+	private IntegrationTestHelper helper = new IntegrationTestHelper();
+
+	@BeforeClass
+	public void setup() {
+
+		client = helper.createClientForBasicAuth();
+		factory = new ResourceFactory(client);
+		OpenshiftProjectRequest projectRequest = factory.create(VERSION, ResourceKind.PROJECT_REQUEST);
```
bjartek: I tried to do this, but I cannot get it to compile if I do it like I think you suggest. This code was copied from another test. I was looking for a projectBuilder but could not find one.

bjartek:
I think i have fixed most of your issues now. Feel free to review again.

jcantrill:
@bjartek LGTM.  Can you rebase and squash your commits before I merge.

bjartek:
Rebased and squashed into 1 commit. Feel free to merge.

jcantrill:
[merge]

PR #245, opened 2016-12-13
[OSJC-242] Add Endpoints

jcantrill:
fixes #242 by adding Endpoints

[test]

PR #249, opened 2016-12-16
Use kind from apigroups

jcantrill:
Depends upon #245.  

This PR:
* Use reflection to find classes based on convention in order to remove type registrations
* Allows registration of kinds if needed
* Fixes broken integration tests

cc @jeffmaury @adietish 

[test]


jcantrill:
[merge]

PR #250, opened 2017-01-14
Orc 248 namespaced urls

### LINKED ISSUE #248: Trying to fetch all deploymentConfigs in all namespaces
I get an error "The api endpoint for kind 'DeploymentConfig' requires a namespace" when i try to fetch deploymentConfigs with the code

`        return client.list(ResourceKind.DEPLOYMENT_CONFIG, "", "affiliation=demo");
`

Namespace is left empty since i want deployment configs in all namespaces that match the given filter query. Ideally the method signature should not have the empty namespace at all here. 

What i want to get here is the following URL

>  https:/localhost:8443/oapi/v1/deploymentconfigs?labelSelector=affiliation%3Ddemo

You can get the same if you do 

> oc get dc --all-namespaces -l "affiliation=demo" --loglevel=9

I have looked in the source and it looks like the URLBuilder does not support building an url for a namespaced resource without a namespace. 

I can fix this if we agree on an approach.

bjartek:
Fixes #248

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/DefaultClient.java
+++ b/src/main/java/com/openshift/internal/restclient/DefaultClient.java
@@ -109,18 +109,29 @@ public class DefaultClient implements IClient, IHttpConstants{
 	}
 
 	@Override
+	public IWatcher watch(IOpenShiftWatchListener listener, String...kinds) {
+		return this.watch("", listener, kinds);
+	}
+
+	@Override
 	public String getResourceURI(IResource resource) {
 		return new URLBuilder(getBaseURL(), typeMapper, resource).build().toString();
 	}
 
 	@Override
 	public <T extends IResource> List<T> list(String kind) {
-		return list(kind,""); //assumes namespace=default
+		return list(kind,"");
 	}
-	
+
+	@Override
+	public <T extends IResource> List<T> list(String kind,  Map<String, String> labels) {
+		return list(kind,"", labels);
+	}
+
+
 	@Override
 	public <T extends IResource> List<T> list(String kind, String namespace) {
-		return list(kind, namespace, new HashMap<String, String>());
+		return list(kind, namespace, new HashMap<>());
 	}
 	
 	@SuppressWarnings("unchecked")
--- a/src/main/java/com/openshift/restclient/IClient.java
+++ b/src/main/java/com/openshift/restclient/IClient.java
@@ -28,16 +28,44 @@ import com.openshift.restclient.model.JSONSerializeable;
  *
  */
 public interface IClient extends ICapable, Cloneable {
-	
+
+	/*
+	 * Watch for changes
+	 *
+	 * @param namespace The namespace to watch for changes in
+	 * @param listener  The listener to be notified on events
+	 * @param kids      The kinds to watch for
+
+	 */
 	IWatcher watch(String namespace, IOpenShiftWatchListener listener, String...kinds);
-	
+
+	/*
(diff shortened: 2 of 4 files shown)


src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -109,18 +109,29 @@ public IWatcher watch(String namespace, IOpenShiftWatchListener listener, String
...
+	}
+
+	@Override
 	public String getResourceURI(IResource resource) {
 		return new URLBuilder(getBaseURL(), typeMapper, resource).build().toString();
 	}
 
 	@Override
 	public <T extends IResource> List<T> list(String kind) {
-		return list(kind,""); //assumes namespace=default
```
bjartek: Does this means that we need to do a major release? If the previous behavior here was that empty namespace meant default.

src/main/java/com/openshift/restclient/IClient.java
```diff
@@ -28,16 +28,44 @@
  *
  */
 public interface IClient extends ICapable, Cloneable {
-	
+
```
bjartek: Javadoc here was missing so i added it

src/test/java/com/openshift/internal/restclient/DefaultClientFilterIntegrationTest.java
```diff
@@ -143,13 +148,28 @@ private static IBuildConfig createBuildConfigWithLabels(IProject project, String
 		IBuildConfig bc = new BuildConfigBuilder(client)
 				.named(name)
 				.inNamespace(project.getNamespace())
```
bjartek: This change somehow was lost when i reabased. Will fix this in a new commit

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -109,18 +109,29 @@ public IWatcher watch(String namespace, IOpenShiftWatchListener listener, String
 	}
 
 	@Override
+	public IWatcher watch(IOpenShiftWatchListener listener, String...kinds) {
+		return this.watch("", listener, kinds);
```
jcantrill: @liggitt I have concerns about allowing users to watch the default namespace.  Doesnt this mean "watch for changes of the given type across all namespaces in the cluster"?  I recall this being 'expensive'

liggitt: it can be, and is unlikely to work for normal users (just like the list below, I guess)

bjartek: We do this at work. We USE the kubernetes-model package and manual http code since it is not supported by this library.

bjartek: This usecase is a totally legitimete usecase that Can be done with the http api manuelly and with the oc client manually. Why should it not be supported? 

What about creating a PriviledgedClient that has these aditional endpoints that needs priviledge?

liggitt: > This usecase is a totally legitimete usecase that canbe done with the http api manually and with the oc client manually. Why should it not be supported?

Agree, just like the list call below

jcantrill:
@bjartek please squash..before merge.  LGTM

bjartek:
Squashed.

jcantrill:
[merge]

PR #252, opened 2017-01-18
allow use of okhttp3 default hostname verifier with cert callback

gabemontero:
@jcantrill PTAL

for the jenkins plugin, we are better served to use the default hostname provider supplied by okhttp3 (i.e. `ISSLCertificateCallback.allowHostname`), but we still want supply logic in the `allowCertificate` path.

All this stems from the discussions with Clayton in https://github.com/openshift/origin/issues/12476

### DIFF
--- a/src/main/java/com/openshift/restclient/ClientBuilder.java
+++ b/src/main/java/com/openshift/restclient/ClientBuilder.java
@@ -46,6 +46,7 @@ public class ClientBuilder {
 	
 	private String baseUrl;
 	private ISSLCertificateCallback sslCertificateCallback = new NoopSSLCertificateCallback();
+	private boolean sslCertCallbackWithDefaultHostnameVerifier = false;
 	private X509Certificate certificate;
 	private String certificateAlias;
 	private IResourceFactory resourceFactory;
@@ -75,6 +76,11 @@ public class ClientBuilder {
 	public ClientBuilder sslCertificateCallback(ISSLCertificateCallback callback) {
 		this.sslCertificateCallback = callback == null ? new NoopSSLCertificateCallback() : callback;
 		return this;
+	}
+	
+	public ClientBuilder sslCertCallbackWithDefaultHostnameVerifier(boolean b) {
+	    this.sslCertCallbackWithDefaultHostnameVerifier = b;
+	    return this;
 	}
 	
 	public ClientBuilder sslCertificate(String alias, X509Certificate cert) {
@@ -190,9 +196,11 @@ public class ClientBuilder {
 				.readTimeout(readTimeout, readTimeoutUnit)
 				.writeTimeout(writeTimeout, writeTimeoutUnit)
 				.connectTimeout(connectTimeout, connectTimeoutUnit)
-				.hostnameVerifier(this.sslCertificateCallback)
 				.sslSocketFactory(sslContext.getSocketFactory(), trustManager);
 
+			if (!this.sslCertCallbackWithDefaultHostnameVerifier)
+			    builder.hostnameVerifier(sslCertificateCallback);
+			
 			if (proxyAuthenticator != null) {
 				builder.proxyAuthenticator(proxyAuthenticator);
 			}


src/main/java/com/openshift/restclient/ClientBuilder.java
```diff
@@ -183,7 +189,9 @@ public IClient build() {
 			dispatcher.setMaxRequests(maxRequests);
 			dispatcher.setMaxRequestsPerHost(maxRequestsPerHost);
 
-			OkHttpClient.Builder builder = new OkHttpClient.Builder()
+			OkHttpClient.Builder builder = null;
+			if (!this.sslCertCallbackWithDefaultHostnameVerifier) {
```
jcantrill: I'm missing the point of the change here.  It looks like the only difference is to alter the read and write timouts  if we turn off the default hostname verifier.  I don't see the relation to between verification and read/writes on the channel.  We should be building the client with the read/write timeouts regardless

jcantrill:
[test]

jcantrill:
Also, do we have any tests we can add to go along with the changes?

src/main/java/com/openshift/restclient/ClientBuilder.java
gabemontero: I'll try to clarify.  The point of the change is the `.hostnameVerifier(this.sslCertificateCallback)` call.  If we make this call, we bypass the use of the default hostname verifier, okhttp3.internal.tls.OkHostnameVerifier, and force implementors to duplicate all the good work okhttp3 did in supporting http://www.ietf.org/rfc/rfc2818.txt and https://tools.ietf.org/html/rfc6125 in the `ISSLCertificateCallback.allowHostname` call.

The jenkins-plugin is fine with letting okhttp3 handle hostname verification.  But we still want to leverage the `ISSLCertificateCallback.allowCertificate` callback when setting up the trust store in com.openshift.restclient.ClientBuilder.getCurrentTrustManager(TrustManagerFactory).

gabemontero:
On testing, I looked at com.openshift.internal.restclient.DefaultClientIntegrationTest and com.openshift.internal.restclient.IntegrationTestHelper and could not infer a clean way to verify things wrt this.

I'll take another pass, but suggestions would be welcome depending on how strongly you want a test here.

src/main/java/com/openshift/restclient/ClientBuilder.java
gabemontero: Also, the default behavior should stay the same ... setting an ISSLCertificateCall should both update the trust store with that callback as well as update the hostname verifier in okhttp3.

gabemontero:
Yeah on the tests, there are not getters on the ClientBuilder, OpenShiftAuthenticator, or OkHttpClient classes to confirm those properties are set.  And without actually initiating an SSL flow with a test specific ISSLCertificateCallback implementation to confirm the callbacks, not sure what to do here.  Note, there are no existing test cases that target ISSLCertificateCallback today.

src/main/java/com/openshift/restclient/ClientBuilder.java
jcantrill: I see now you are doing `builder.hostnameVerifier(this.sslCertificateCallback)`  or not.  I would suggest:

* leaving the code as is instead of duplicating a good portion of it
* and an 'if' block after the builder assignment that evaluates setting calling `builder.hostnameVerifier(this.sslCertificateCallback)`

gabemontero: duh ... for some reason I thought the `build()` was done as part of all that, but now I see it is on the next line.

will push an update momentarily

gabemontero: update pushed

gabemontero:
hmmm ... the space/tabbing complicated the github diff a little ... i'll see if i can address

gabemontero:
yeah the github diff is happier now

jcantrill:
[merge]

gabemontero:
thanks @jcantrill - also, when you get the chance, per the discussion in the other PR, etc. please push to the maven repo and cut a final release.  thanks again

PR #258, opened 2017-02-11
POC of async client

bjartek:
This is not meant to be merged in right now. Just want to show what I have been trying out. 

The problem that I see here is that DefaultClient really should return the Call. Not handle the call syncronously. Could be abstract DefaultClient into something like a BaseClient that returns the Call and then make Default client handle the call syncronously and AsyncClient handle it async?


bjartek:
I did not find time to work on this, and it is outdated now so i will just close.

PR #277, opened 2017-07-03
jbide-23862: User should be able to see what OpenShift version a connection is

bdshadow:
@jcantrill @jeffmaury @adietish please, take a look. I've changed the api of the `IClient`. Is it ok? Do you know anybody else who uses this rest client, except us? Will it be critical for them? If yes, there is no problem to leave `IClient` as it is and add the new methods only to the `DefaultClient` class

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/DefaultClient.java
+++ b/src/main/java/com/openshift/internal/restclient/DefaultClient.java
@@ -20,6 +20,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Consumer;
 
 import org.apache.commons.lang.ObjectUtils;
 import org.apache.commons.lang.StringUtils;
@@ -46,6 +47,8 @@ import com.openshift.restclient.model.IList;
 import com.openshift.restclient.model.IResource;
 import com.openshift.restclient.model.JSONSerializeable;
 
+import okhttp3.Call;
+import okhttp3.Callback;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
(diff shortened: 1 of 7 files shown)


bdshadow:
don't merge! Found some issue.
Tests are failing + for inaccessible openshift server, the client hangs

bdshadow:
tests are fixed.
Version fetch made non-blocking. Can be reviewed, testes and merged

src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -81,8 +84,8 @@ public DefaultClient(URL baseUrl,  OkHttpClient client, IResourceFactory factory
 		if(this.factory != null) {
 			this.factory.setClient(this);
 		}
-		openShiftVersion = System.getProperty(SYSTEM_PROP_OPENSHIFT_API_VERSION, null);
-		kubernetesVersion = System.getProperty(SYSTEM_PROP_K8E_API_VERSION, null);
+		initMasterVersion("version/openshift", new VersionCallback(version -> this.openShiftVersion = version));
```
jcantrill: Before changing these, please ensure they are not used in some other capacity. There original intention represented API version of the endpoints and not version of the overall product.

bdshadow: @jcantrill `openShiftVersion` and `kubernetesVersion` are used nowhere, except `DafaultClient`. Moreover, it doesn't have sense to be these variables responsible for API version, because it doesn't really provide any useful information. It is always v1 (agree, that some long time ago it was v1beta). Plus, if it is API versions why are 2 variables used then? Anyway the versions are the same - look at `OpenShiftAPIVersion` and `KubernetesAPIVersion` enums - they're exact copies of each others

bdshadow:
@jcantrill aren't commits squashed during the merge? In jbosstools-openshift we have a "Squash and merge" button for PRs. Can't find it here since i don't have access.

bdshadow:
sorry, closed it because of the wrong branch name and wrong jira numbers in the commit comments. See https://github.com/openshift/openshift-restclient-java/pull/278 now

PR #280, opened 2017-07-20
add restclient prefix to normal okhttp setting of user-agent

gabemontero:
@jcantrill - during some debug with the struggles on us-east prod online offering, some of the analysis pointed to the "okhttp" client creating an abundance of watches.  By default okhttp uses its Version class to set the User-Agent.

Of course, there are multiple users of okhttp under the openshift/eclispe/fabric umbrella :-)

I've started updating some of the relevant repos (at least the ones our Jenkins image brings in) to get a little more specific wrt setting the User-Agent, following an approach that fabric employs.

ptal

@bparees fyi

### DIFF
--- a/src/main/java/com/openshift/restclient/ClientBuilder.java
+++ b/src/main/java/com/openshift/restclient/ClientBuilder.java
@@ -26,6 +26,8 @@ import javax.net.ssl.TrustManager;
 import javax.net.ssl.TrustManagerFactory;
 import javax.net.ssl.X509TrustManager;
 
+import org.apache.commons.lang.StringUtils;
+
 import com.openshift.internal.restclient.DefaultClient;
 import com.openshift.internal.restclient.ResourceFactory;
 import com.openshift.internal.restclient.authorization.AuthorizationContext;
@@ -36,7 +38,11 @@ import com.openshift.restclient.utils.SSLUtils;
 
 import okhttp3.Authenticator;
 import okhttp3.Dispatcher;
+import okhttp3.Interceptor;
 import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+import okhttp3.internal.Version;
 
 /**
  * Builder to create IClient instances.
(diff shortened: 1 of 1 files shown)


src/main/java/com/openshift/restclient/ClientBuilder.java
```diff
@@ -199,6 +203,13 @@ public IClient build() {
 
 			OkHttpClient.Builder builder = new OkHttpClient.Builder()
 				.addInterceptor(responseCodeInterceptor)
+				.addNetworkInterceptor(new Interceptor() {
+                    @Override
+                    public Response intercept(Chain chain) throws IOException {
+                        Request agent = chain.request().newBuilder().header("User-Agent", "openshift-restclient-java/"+Version.userAgent()).build();
```
jcantrill: Might it be more useful to allow a user of the ClientBuilder to set the useragent so that you could determine from which application requests are originating?  Isn't that part of the problem we are trying to address here, who is the bad actor?

gabemontero: Yeah I was in a quick / dirty frame of mind when first crafting this, but the pressure on this has decreased a bit.  I still want to include some notion of restclient/fabric version, but I'll change so the ClientBuilder consumer can prepend how they are.

Thanks.

gabemontero: update pushed

src/main/java/com/openshift/restclient/ClientBuilder.java
```diff
@@ -196,9 +206,17 @@ public IClient build() {
 			//if we need to really expose them.
 			dispatcher.setMaxRequests(maxRequests);
 			dispatcher.setMaxRequestsPerHost(maxRequestsPerHost);
+			final String userAgent = (this.userAgentPrefix != null ? this.userAgentPrefix + "/" : "") + "openshift-restclient-java/"+Version.userAgent();
```
jcantrill: There are some useful utilities from the Apache deps StringUtils Objectutils which could make this prettier.

gabemontero: StringUtils update pushed

jcantrill:
[test]

jcantrill:
[merge]

PR #291, opened 2017-09-27
revert PR 284

gabemontero:
@jcantrill - still not sure how my prior openshift testing missed this, but I've now confirmed that bumping okhttp is a more invasive change.

Specifically, it appears that okhttp-ws has been deprecated (see [1], [2]) and replaced with native websocket support in okhttp3 around v3.5

The internal classes that okhttp-ws depend on no longer exist (I finally figured this out when sorting thorugh  [3]).

While similar, the replacement classes were different enough that switching to the new APIs was not a quick fix for me (and I won't have the cycles to complete it on the side).

Hence, seemed more prudent to revert https://github.com/openshift/openshift-restclient-java/commit/1b29d03c88c8b13133ad5814a11f84b87a46ef2e

restclient integration tests look OK.

apologies for the churn ... assuming nobody else has seen this, and hopefully nobody has been negatively impacted, given the lack of issues, etc.

[1] https://github.com/csync/csync-android/issues/22
[2] https://medium.com/square-corner-blog/web-sockets-now-shipping-in-okhttp-3-5-463a9eec82d1
[3] https://github.com/openshift/jenkins-plugin/pull/149

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -131,12 +131,12 @@
 		<dependency>
 		  <groupId>com.squareup.okhttp3</groupId>
 		  <artifactId>okhttp</artifactId>
-		  <version>3.8.1</version>
+		  <version>3.3.1</version>
 		</dependency>
 		<dependency>
 		  <groupId>com.squareup.okhttp3</groupId>
 		  <artifactId>okhttp-ws</artifactId>
-		  <version>3.4.2</version>
+		  <version>3.3.1</version>
 		</dependency>
 		<dependency>
 			<groupId>org.yaml</groupId>
--- a/src/test/java/com/openshift/internal/restclient/TypeMapperFixture.java
+++ b/src/test/java/com/openshift/internal/restclient/TypeMapperFixture.java
@@ -86,7 +86,6 @@ public class TypeMapperFixture {
 				.request(new Request.Builder().url("https://someurlfortesting").build())
 				.protocol(Protocol.HTTP_1_1)
 				.code(IHttpConstants.STATUS_OK)
-				.message("foo")
 				.body(ResponseBody.create(null, response))
 				.build();
 	}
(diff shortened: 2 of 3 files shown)


adietish:
@gabemontero @jcantrill reading this makes me think that the watch client would be affected? I didnt notice any effect though?

gabemontero:
hey @adietish - apologies, but I'll need a little clarification on what you mean by "watch client" in order to sufficiently answer with any certainty.

What I can say with certainty is that the call in `com.openshift.internal.restclient.api.capabilities.PodExec.start` to `okhttp3.ws.WebSocketCall.enqueue` is the crux of the situation here.  `okhttp3.ws.WebSocketCall.enqueue` is in the `okhttp-ws` dependency, which per the referenced links, is effectively deprecated.  `okhttp3.ws.WebSocketCall.enqueue` in turn was depending on the existence of `okhttp3.internal.Internal.setCallWebSocket` in the `okhttp3` dependency.

However, `okhttp3.internal.Internal.setCallWebSocket` no longer exists in the more recent versions of `okhttp3`.

I only ran into this when trying out a code snippet along the following lines in the openshift jenkins plugin:
```
        IPod p;
        // omitted code that looks up / initialized p
        IStoppable exec = p.accept(
                new CapabilityVisitor<IPodExec, IStoppable>() {
                    @Override
                    public IStoppable visit(IPodExec capability) {
                        return capability.start(
                                new IPodExec.IPodExecOutputListener() {

```

adietish:
@gabemontero thx for the insights. we currently dont use PodExec, so at least in this regard we're not affected. The only part where we're relying on web sockets is afaik com.openshift.internal.restclient.okhttp.WatchClient which doesnt seem to be affected by the above. Correct?

gabemontero:
ah - that "watch client" :-) .... thanks for the clarification @adietish 

Taking a quick peek at that file, it has `import okhttp3.ws.WebSocketCall`

You probably were already aware, but that is a `okhttp-ws` class.  So *in theory* it could be impacted by future upgrades to `okhttp`.  That said, the use of it in `com.openshift.internal.restclient.okhttp.WatchClient` *seems* pretty minimal.  So I suspect the `WatchClient` use of it would still be OK.  But it might be worth experimenting and seeing if you all can just get rid of it.

adietish:
@gabemontero ok, thanks for the insights, sounds like we're safe for now, but might have to go through changes in the future.

Issue #300, opened 2017-11-03
[bug]Rsync is not working when the local folder path to rsync contains a space


apupier:
the returned error is "only SOURCE_DIR and POD:DESTINATION_DIR should be specified as arguments"

apupier:
I hope that this test will emphasize the issue: https://github.com/apupier/openshift-restclient-java/tree/300-RsyncErrorWhenPathContainsSpace

jeffmaury:
I think this is an oc issue. @bdshadow can you confirm ?

bdshadow:
@apupier @jeffmaury most probably it's bug, which we'already discussed with you @apupier . Even though it has been fixed in the upstream oc client, it is backported to old version. So it'll work only starting from some version of oc client (at least 3.6.x must work fine)

apupier:
if you are mentioning this bug https://github.com/openshift/origin/issues/15187 this is not the same.
The pointed bug it is when the "path to oc" has space to it, not when the folder to rsync contains space.

apupier:
I managed to have the Integration test running locally using the IDE (not from maven command-line) but it was enough to emphasize the bug with the test. I updated the branch:
https://github.com/apupier/openshift-restclient-java/tree/300-RsyncErrorWhenPathContainsSpace

![image](https://user-images.githubusercontent.com/1105127/32503055-8a48b03e-c3dc-11e7-8e72-8ba47e37be10.png)

apupier:
I'm able to have it working from cygwin terminal:
```
Aurelien Pupier@DESKTOP-9NT300B /cygdrive/c/Users/AURELI~1/AppData/Local/Temp/junit3759284092951894034
$ oc rsync --token=NvR3P3C2l-r-VQo4dwDAiH23iR4r_rBUgceoWu6ujBY --server=https://192.168.1.39:8443 --insecure-skip-tls-verify=true "C:/Users/Aurelien Pupier/AppData/Local/Temp/junit3759284092951894034/with sp@cé in path" docker-registry-1-9w80n:/tmp -n default
Forwarding from 127.0.0.1:63546 -> 59441
Forwarding from [::1]:63546 -> 59441
Handling connection for 63546
sending incremental file list

sent 142 bytes  received 16 bytes  316.00 bytes/sec
total size is 0  speedup is 0.00
```

but not from cmd:
```
[...]

PR #304, opened 2017-11-09
#300 - Support spaces in paths to synchronize

apupier:
Signed-off-by: Aurélien Pupier <apupier@redhat.com>

Several points:
- I use a regex to parse a String to a List<String>, I think that it would be proper that buildArgs methods will return a List<String> directly (but requires to break at least internal API)
- the addition of double-quotes is not at the same place for Local and Pod Peers, I don't find it very nice but  not sure what is the best way:
-- introduce a new method getCommandLineLocation which will have the double-quote ta the right place?
-- always return location/parameter of Peer with the double-quotes? Is it used somewhere else?

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/capability/resources/AbstractOpenShiftBinaryCapability.java
+++ b/src/main/java/com/openshift/internal/restclient/capability/resources/AbstractOpenShiftBinaryCapability.java
@@ -15,6 +15,8 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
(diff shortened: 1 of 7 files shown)


jeffmaury:
If you returns a List you don't have to add double quotes as ProcessBuilder take care of passing each argument separatetly

src/main/java/com/openshift/internal/restclient/capability/resources/AbstractOpenShiftBinaryCapability.java
```diff
@@ -191,6 +193,15 @@ private ProcessBuilder initProcessBuilder(String location, final OpenShiftBinary
 		return builder;
 	}
 	
+	private List<String> splitTakingCareOfSpaceInsiqueQuotes(String str){
```
jcantrill: Is there a misspelling here? Should it be `splitTakingCareOfSpaceInsideQuotes' .  Can we just call it 'split' since that is what it is doing or 'normalize' or 'sanitize'

apupier: yes, mispelling.
I would prefer to have a more meaningful than just split/normalize/sanitize

apupier:
I tried with a List of Arguments directly instead of creating a String and then reparsing to convert to a List<String>

I have 2 issues currently:
- having a space in target folder is not working (test written which emphasized it)
- Even without space it is not working inside Jboss Tools (not found any reasons)

There is an API break with this implementation, not found a way to not break it, it is for IRsyncable.
It is possible if we increase the library version? or too many clients are relying on it?

> If you returns a List you don't have to add double quotes as ProcessBuilder take care of passing each argument separatetly

i tried without quotes with the List and it is not working. There is a specific case for remote pod, the double-quotes are not around the full argument, it is only on a subpart of it, for instance camel-ose-springboot-xml-2-31gg0:"/deployments/"

jeffmaury:
What I meant is that from the rest client POV, if we managed arguments through a List and because we are using ProcessBuilder then we should be fine. Then if errors kept happening, this is an upstream (oc) issue

apupier:
> What I meant is that from the rest client POV, if we managed arguments through a List and because we are using ProcessBuilder then we should be fine. Then if errors kept happening, this is an upstream (oc) issue

Being the devil advocate: if oc rsync requires to put quotes around only the path part, the answer will be that "it is how it is working" and that you can implement the "new feature" if you want

apupier:
> There is an API break with this implementation, not found a way to not break it, it is for IRsyncable.
> It is possible if we increase the library version? or too many clients are relying on it?

up

jeffmaury:
Replaced by #310

Issue #318, opened 2018-03-07
OpenShiftBinaryRSync should allow several exclusion filters (ex. .git, .npm)


adietish:
when executing a OpenShiftBinaryRSync one should be able to provide several exclusion filters, no just the exclude for .git that we support currently

adietish:
deployed maven artifact to https://repository.jboss.org/nexus/service/local/repositories/snapshots/content/com/openshift/openshift-restclient-java/6.0.0-SNAPSHOT/openshift-restclient-java-6.0.0-20180309.154825-1.jar

PR #319, opened 2018-03-07
[318] Multiple excludes for binary rsync

adietish:
provides a fix for #318 

This is breaking the public API, thus it's suggested at 6.0.0-SNAPSHOT

### DIFF
--- a/src/main/java/com/openshift/restclient/capability/IBinaryCapability.java
+++ b/src/main/java/com/openshift/restclient/capability/IBinaryCapability.java
@@ -12,21 +12,24 @@ package com.openshift.restclient.capability;
  * @author Andre Dietisheim
  */
 public interface IBinaryCapability extends ICapability {
-	
+
+	static final OpenShiftBinaryOption SKIP_TLS_VERIFY = new SkipTlsVerify();
+
 	/**
-	 * Optional arguments to pass when running the {@code oc} command.
+	 * Skips the SSL/TLS Verification when using a binary capability
 	 */
-	public enum OpenShiftBinaryOption {
-		/** option to skip verifying the certificates during TLS connection establishment. */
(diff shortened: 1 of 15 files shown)


src/main/java/com/openshift/restclient/capability/IBinaryCapability.java
```diff
@@ -12,21 +12,24 @@
...
+		@Override
+		public void append(StringBuilder commandLine) {
+			commandLine.append(" --insecure-skip-tls-verify=true");
+		}
 	}
 	
+	abstract static class CommandLineArgument {		
+		public abstract void append(StringBuilder builder);
+	}
+
```
jeffmaury: Why not an interface ?

jeffmaury:
This is a breaking change so I would favor switch to 6.x version. Also there are a couple of artifacts using this in addition to JBossTools (see https://mvnrepository.com/artifact/com.openshift/openshift-restclient-java/usages), I checked openshift-pipeline don't seem to be affected but maybe worth checking

adietish:
@jeffmaury @jcantrill this is ready for merge, please review

src/main/java/com/openshift/restclient/capability/IBinaryCapability.java
jeffmaury: And having it a functional interface would allow lamba to be used

adietish:
[test]

adietish:
once this PR is approved we should release a 5.9.6.Final which has the fix for ignoring .git folders in OpenShiftBinaryRSync since this PR is bumping openshift-restclient-java to 6.0.0-SNAPSHOT

adietish:
@jcantrill thx!

adietish:
[merge]

Issue #326, opened 2018-04-03
Allow easy extension of missing domain objects.


Tanemahuta:
When encountering missing domain objects, we always need to create a pull request and wait for a release. We would appreciate, if we could extend the domain objects and mappings ourself.
This could be done via a registry, for example.

PR #327, opened 2018-04-03
Enhanced code the way that ppl can actually implement their own resou…

Tanemahuta:
…rce kinds.

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
+++ b/src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
@@ -4,28 +4,11 @@
  * This program is made available under the terms of the 
  * Eclipse Public License v1.0 which accompanies this distribution, 
  * and is available at http://www.eclipse.org/legal/epl-v10.html 
- * 
+ *
  * Contributors: 
  * Red Hat, Inc. - initial API and implementation 
  ******************************************************************************/
 package com.openshift.internal.restclient;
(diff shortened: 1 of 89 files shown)


Tanemahuta:
This is for issue #326 326

jcantrill:
I'm not sure what actually the changes are here since a number of them appear to be reformatting. @adietish we should really have some lint or other that  requires spaces instead tabs.  Additionally, I am unlikely to accept this PR given I cant see what the actually changes are; the diff is too great

Tanemahuta:
Well, provide your formatting options, and I run them again through my IDE...
I just wanted to provide a way to extend your API in case a domain object does not yet exists through that registry in the client builder. This would improve extending the client dramatically, since everyone could try out their extensions and then integrate them easily through the PredefinedResourceKinds.
Also there might be an option to diff ignoring whitespaces.

adietish:
@Tanemahuta great idea really! But as @jcantrill pointed out reviewing your PR is very hard since the significant changes are buried by the reformatting noise. We'd appreciate a lot if you could separate the functional changes from the formatting to facilitate our task.

jcantrill:
@Tanemahuta Just merged changes that include reformatting and check style.  Please rebase off those changes so we can get to the meat of what you are proposing

Tanemahuta:
I'm on it right now. Will issue a new pull request rebased from master.

PR #329, opened 2018-04-20
Add Proxy/ProxySelector capabilities to ClientBuilder

adietish:


vladfau:
@adietish, added some tests which verify that Proxy/ProxySelector is set and match expected ones and not set (or used defaults) in the other case. I'm not really familiar with stubbing/mocking in Java as I'm not really proficient in Java testing, but I used reflection to get `proxy`/`proxySelector` fields from `IClient`. 

We can either leave it as is or create getters.

src/test/java/com/openshift/internal/restclient/okhttp/ClientProxyConnectionTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+		OkHttpClient okClient = getOkHttpClient(client);
+		ProxySelector proxySelector = okClient.proxySelector();
+
+		assertEquals(proxySelector, ProxySelector.getDefault());
+	}
+
+
+	private OkHttpClient getOkHttpClient(IClient client) throws NoSuchFieldException, IllegalAccessException {
+		Field clientField = client.getClass().getDeclaredField("client");
```
jcantrill: Any reason to not cast this to the impl of IClient, return the value directly instead of using reflection?  What advantage here is there in using reflection?

vladfau: Third option is to make OkHttpClient protected. WDYT?

vladfau: @jcantrill, is it better now?

jcantrill:
@vladfau Please rebase

src/test/java/com/openshift/internal/restclient/okhttp/ClientProxyConnectionTest.java
jcantrill: yes

vladfau:
@jcantrill @adietish rebased and ready to go

jcantrill:
@vladfau LGTM.  Please squash and I we can merge

jcantrill:
[merge]

PR #330, opened 2018-04-20
Add checkstyle and fix violations

jcantrill:
This PR adds in checkstyle and fixes all violations associated with it, including reformating to remove tab chars.  It adds or changes no functionality.  This should help us better resolve #327


jcantrill:
@jeffmaury We can set the styles to any we prefer.  Its defined by `checkstyle.xml` which is customizable or updatable.  I'll merge this and defer to you to change it as you desire.

PR #335, opened 2018-06-04
326: Allow easy extension of missing domain objects.

Tanemahuta:
There you go. I've reformatted it.

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/DefaultClient.java
+++ b/src/main/java/com/openshift/internal/restclient/DefaultClient.java
@@ -21,6 +21,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Consumer;
 
 import org.apache.commons.lang.ObjectUtils;
@@ -31,13 +32,15 @@ import org.slf4j.LoggerFactory;
 
 import com.openshift.internal.restclient.authorization.AuthorizationContext;
 import com.openshift.internal.restclient.okhttp.WatchClient;
+import com.openshift.restclient.DefaultResourceKindRegistry;
 import com.openshift.restclient.IApiTypeMapper;
 import com.openshift.restclient.IClient;
 import com.openshift.restclient.IOpenShiftWatchListener;
 import com.openshift.restclient.IResourceFactory;
 import com.openshift.restclient.IWatcher;
 import com.openshift.restclient.OpenShiftException;
-import com.openshift.restclient.ResourceKind;
+import com.openshift.restclient.PredefinedResourceKind;
+import com.openshift.restclient.ResourceKindRegistry;
 import com.openshift.restclient.UnsupportedOperationException;
 import com.openshift.restclient.api.ITypeFactory;
 import com.openshift.restclient.authorization.IAuthorizationContext;
@@ -118,7 +121,7 @@ public class DefaultClient implements IClient, IHttpConstants {
 
     @Override
     public String getResourceURI(IResource resource) {
-        return new URLBuilder(getBaseURL(), typeMapper, resource).build().toString();
+        return new URLBuilder(getBaseURL(), typeMapper, resource, getResourceKindRegistry()).build().toString();
     }
 
     @Override
@@ -236,10 +239,9 @@ public class DefaultClient implements IClient, IHttpConstants {
             params = Collections.emptyMap();
         }
 
-        if (ResourceKind.LIST.equals(kind)) {
+        if (PredefinedResourceKind.LIST.equals(kind))
             throw new UnsupportedOperationException("Generic create operation not supported for resource type 'List'");
-        }
-        final URL endpoint = new URLBuilder(this.baseUrl, typeMapper).kind(kind).name(name).namespace(namespace)
+        final URL endpoint = new URLBuilder(this.baseUrl, typeMapper, getResourceKindRegistry()).kind(kind).name(name).namespace(namespace)
                 .subresource(subresource).subContext(subContext).addParameters(params).build();
 
         try {
@@ -494,4 +496,9 @@ public class DefaultClient implements IClient, IHttpConstants {
         return null;
     }
 
+    protected ResourceKindRegistry getResourceKindRegistry() {
+        return Optional.ofNullable(factory)
+                .map(IResourceFactory::getResourceKindRegistry)
+                .orElseGet(DefaultResourceKindRegistry::new);
+    }
 }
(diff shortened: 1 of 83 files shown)


Obirah:
Great idea, thank you so much!

adietish:
[test]

adietish:
@Tanemahuta this looks awesome at first sight. This being quite a big PR can you please point out what you changed?

Tanemahuta:
Hello there, I introduced a registry for the ResourceKinds and their implementations,
so whenever you factor a Client, you can now provide your own Registry which can be used to register new Implementations for ResourceKinds which have not been defined yet in the connector.
This way folks can instantly work on implementing those new Resources, try them out and then create a PR (and thus do not have to wait for the new Release).

adietish:
@Tanemahuta sounds great, thx for contributing!
The first thing that I see that I think needs discussing is your renaming ResourceKind -> PredefinedResourceKind. This is a breaking change which we should either get away from (to keep backward compatibility) or change the librarys major version.
I didnt see unit tests for adding/removing resource kinds. Having those is highly encouraged since this provides safeguards for eventually evolving the library and would also help user-programmers to find out how to achieve this.

Tanemahuta:
You are right, I haven't written any, will do in the next few days, promised.

adietish:
@Tanemahuta great, thx!

jcantrill:
I would add to this renaming is there is a concept in Kubernetes of CustomResourceDefinition with which I think this rename conflicts.

adietish:
@jcantrill I read https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/ and see that apparently Kubernetes knows the concept of a custom resource. What this PR tries to achieve is an "unknown" resource as in "standard resource but not know to the library". So I see a problem in where the PredefinedResourceType might be mistakenly taken for the concept of non-custom resources in kubernetes. Can you elaborate on where you see a clash?

jcantrill:
@adietish Yes, I understand that.  I think "PredefinedResourceKind" is just not a good name.  I think "ResourceKind" is better and there maybe should be a concept of 'ResourceKindExtensions' or custom or something that:
1. Is not such a large, disruptive change to the code base
1. Is a better indicator that you the developer are providing something that is not part of the library.

adietish:
@jcantrill +1

adietish:
@Tanemahuta you seem not to agree. Arguments?

Tanemahuta:
Look, I do not care about your naming style here, I just want this to be done, so everyone can enjoy the client. I would suggest the following things:
a) you as the developer team agrees on a name, suggestion: 
 - keep ResourceKind as the interface, though this breaks the API and introduces a Major version change, but as you can see from the code, this is a necessary evil
 - Rename the enum PredefinedResourceKind to ImplementedResourceKind or something alike
b) after you have agreed, I will change my commit, add a unit test and push it.

I just want to say, that it doesn't really help if you start tearing apart PRs with "potential naming clashes ". From the code I have seen in the project, I'd say this is the least of the problems. And if there's really an issue with naming, please just suggest a naming pattern. Thank you.

jcantrill:
`<soapbox>`
@Tanemahuta Let's leave the snarkyness for your visits [abroad](https://goo.gl/j6Zv7F).  The fact is you are renaming a class that is used by other people besides you.  The ramifications are that anyone consuming this client with your PR merged with have to refactor and absorb it; you have provided no backwards compatibility.  I can appreciate your frustrations, but the primary consumer of this plugin is the Eclipse Openshift plugin which is a team of limited resources and frankly does not have the time to refactor the the code in a timely manor while still meeting their other timelines and release objectives.

Additionally, the original implementation of ResourceKind was purposely not an enum for the shear reason we understand, by experience, the errors that occur when the enum doesn't have a value that gets deserialized from the original json.  It was purposely left as a string so the client could handle causes of kinds it doesn't explicitly know about.

Now if we can all be collaborative adults....
`</soapbox>`

* Leave `com.openshift.restclient.ResourceKind` and add a deprecation notice
* Introduce `com.openshift.restclient.api.ResourceKind` to replace the former
* provide an adapter from one to the other or maybe some way to get results for 'api.ResourceKind' from a 'ResourceKind'

These are just off the top of my head so I have no understanding of the difficulty or feasibility.

Note:  
* Ideally we desire to move to a pattern of generating the models like https://github.com/fabric8io/kubernetes-client or even completely deprecating this client in favor of that one.
* We know this code needs some love [love](https://github.com/openshift/openshift-restclient-java/pull/335/files#diff-097ae2a6486af7c7fd6c150885cfb7b0R9) and you're welcome to submit prs

Tanemahuta:
Jeff, I'm being an adult. Adult is also based on proper communication: Criticism without proper explanation and/or improvement proposals are rarely being understood, am I right or not?

Now back to the technical topic: refactoring is not a big deal here (all you would need to do is add a .getIdentifier()) for the resource types. That being said, and you bein' heard, I propose the following change: 
- ResourceKind stays as it is in the original code, and the string representations of the constants are being used
- Rename that Registry into ModelRegistry
- the "ModelRegistry" will now be used with a String->ImplClass pattern, which can be used to register/override the implementation classes based on the identifier
That should reduce refactoring issued to nearly zero

Agree?

jcantrill:
> Now back to the technical topic: refactoring is not a big deal here (all you would need to do is add a .getIdentifier()) for the resource types. That being said, and you bein' heard, I propose the following change:

Seems reasonable. I actually prefer using enums but I don't believe there is a way to gracefully fallback for the case where and unexpected value is deserialzed; I don't think there is a choice here.

>ResourceKind stays as it is in the original code, and the string representations of the constants are being used
>Rename that Registry into ModelRegistry

Seems reasonable

>the "ModelRegistry" will now be used with a String->ImplClass pattern, which can be used to >register/override the implementation classes based on the identifier

Seems resonable

adietish:
@jcantrill @Tanemahuta happy that we get back to an objective tone discussing code rather than calling names which doesn't help.

So yeah, the compromise that you found sounds reasonable to me, too.

I didnt spot the enum that was brought up in the discussion further up (did i miss something?).
For the background on why one should use enums only for immutable alternatives:
imagine outdated client v1.0, which tries to talk to server v2.0 which introduced new resources. Client 1.0, which tries to parse an unknown resource will end up throw IllegalArgumentException rather than handle this using something like an UNKNOWN type (one can override enum#valueOf to a more lenient variant, but #valueOf is still there and too tempting to use). 
And then, for the background again: backward compatibility IS a major concern for a library which is used by a range of consumers and for a prolonged time. I personally prefer breaking changes if it's for the sake of code clarity. But still, one needs to ponder benefits/drawbacks carefully.

PR #342, opened 2018-08-07
Support base-10 units for PV sizes

davidgubler:
I've tried to add support for base-10 units without introducing incompatibilities. Note, however, that mixing base-10 and base-2 units in a cluster basically makes the PersistentVolume.getCapacity(String) and PersistentVolume.getCapacity(MemoryUnit) methods useless, because they return 0 if conversion between the two bases would be required (I'm not saying this makes sense, I just tried to change the behavior as little as possible). IMHO these methods should be deprecated and eventually removed. getCapacity() is fine because it returns bytes, which always works.

What do you think?

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/model/volume/PersistentVolume.java
+++ b/src/main/java/com/openshift/internal/restclient/model/volume/PersistentVolume.java
@@ -18,6 +18,8 @@ import java.util.regex.Pattern;
 
 import org.jboss.dmr.ModelNode;
 import org.jboss.dmr.ModelType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.openshift.internal.restclient.model.KubernetesResource;
 import com.openshift.internal.restclient.model.volume.property.HostPathVolumeProperties;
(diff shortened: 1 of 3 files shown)


src/main/java/com/openshift/internal/restclient/model/volume/PersistentVolume.java
```diff
@@ -49,22 +49,22 @@ public PersistentVolume(ModelNode node, IClient client, Map<String, String[]> pr
...
      * @return 0, if conversion not possible. Otherwise the capacity in given units.
      * @see {@link PersistentVolume#convert(String, MemoryUnit)}
      * @throws IllegalArgumentException
      *             if not supported {@code unit} given.
      */
     @Override
     public long getCapacity(String unit) {
-        String capacity = asString(PV_CAPACITY);
-        return convert(capacity, MemoryUnit.valueOf(unit));
+        return getCapacity(MemoryUnit.valueOf(unit));
```
jcantrill: @adietish We've made a concerted effort to stay away from enums to make this client lenient.  Seems like here is a ase that's acceptible.

src/main/java/com/openshift/internal/restclient/model/volume/PersistentVolume.java
```diff
@@ -227,40 +227,30 @@ private IHostPathVolumeProperties createHostPathVolumeProperties() {
...
         MemoryUnit currentUnit = parseCapacityUnit(capacity);
 
-        int operation = currentUnit.compareTo(designatedUnit);
-        if (operation == 0L) {
-            return number;
-        } else if (operation > 0L) {
-            return safeMultiplicationBy1024(number, operation); // number << (10 * operation);
-        } else {
-            return 0L;
+        long bytes = Math.multiplyExact(number, currentUnit.getFactor());
```
jcantrill: Probably should be some log warning message here identifying the value was not convertable

adietish:
@davidgubler I'd love to merge this in if you can address the suggestion that @jcantrill made. Can you comment & address those?

srueg:
@adietish any chance we can get this in?

PR #352, opened 2018-10-22
nodePort enhancement

### LINKED ISSUE #235: IServicePort does not include a method to get/set the Node port
When calling Service.getPort() the IServicePort interface only returns the port and targetPort.  There is no interface to return the nodePort.

Stealth-ip:
added setter and getter for nodePort
tests changed

fixes #235

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/model/ServicePort.java
+++ b/src/main/java/com/openshift/internal/restclient/model/ServicePort.java
@@ -28,6 +28,7 @@ public class ServicePort extends ModelNodeAdapter implements IServicePort {
     private static final String PROPERTY_PORT = "port";
     private static final String PROPERTY_PROTOCOL = "protocol";
     private static final String PROPERTY_TARGET_PORT = "targetPort";
+    private static final String PROPERTY_NODE_PORT = "nodePort";
     private static final Map<String, String[]> KEY_MAP = new HashMap<>();
 
     static {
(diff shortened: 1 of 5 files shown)


adietish:
@jcantrill containers also have ports. Do I get it right that the nodePort is only available for services (that's what the scarce documentation that I found seems to imply)?
I'm asking because this change caused a compilation error in JBT for (ServicePortAdapter)[https://github.com/jbosstools/jbosstools-openshift/blob/master/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicePortAdapter.java#L23]). This adapter can be initialized by an (IServicePort ) [https://github.com/openshift/openshift-restclient-java/blob/master/src/main/java/com/openshift/restclient/model/IServicePort.java] which holds the nodePort, but also by an (IPort)[https://github.com/openshift/openshift-restclient-java/blob/master/src/main/java/com/openshift/restclient/model/IPort.java]  which lacks it.

jcantrill:
> @jcantrill Do I get it right that the nodePort is only available for services (that's what the scarce documentation that I found seems to imply)? In other terms, [IPort](https://github.com/openshift/openshift-restclient-java/blob/master/src/main/java/com/openshift/restclient/model/IPort.java) doesn't need any update?
> 
@adietish 
From `k8s.io/kubernetes/staging/src/k8s.io/api/core/v1/types.go` that appears to be correct which would make sense.  "Port" is an abstraction that allows ports to be named on services and pods.  Ports for a cluster node can also be fronted by a service which wouldn't make sense in another context.

> I'm asking because this change caused a compilation error in JBT for [ServicePortAdapter](https://github.com/jbosstools/jbosstools-openshift/blob/master/plugins/org.jboss.tools.openshift.ui/src/org/jboss/tools/openshift/internal/ui/wizard/deployimage/ServicePortAdapter.java#L23). This adapter can be initialized by an [IServicePort ](https://github.com/openshift/openshift-restclient-java/blob/master/src/main/java/com/openshift/restclient/model/IServicePort.java) which holds the nodePort, but also by an [IPort](https://github.com/openshift/openshift-restclient-java/blob/master/src/main/java/com/openshift/restclient/model/IPort.java) which lacks it.

I allowed the merge because I thought the tests all succeed.  We should revert this PR and ask @Stealth-ip to resubmit.

adietish:
@jcantrill thx for confirming. I corrected things in JBT and we're all fine. I just wanted to make sure that nothing went forgotten here.

Issue #357, opened 2018-12-07
does not run with java 11


adietish:
given that java 10 went EOF we should now make sure openshift-restclient-java runs with java 11.
Currently the following exception occurs:
```

java.lang.NoClassDefFoundError: javax/xml/bind/DatatypeConverter
--
at com.openshift.restclient.utils.Base64Coder.encode(Base64Coder.java:43)
at com.openshift.restclient.utils.Base64Coder.encode(Base64Coder.java:55)
at com.openshift.internal.restclient.okhttp.BasicChallangeHandler.handleChallange(BasicChallangeHandler.java:47)
at com.openshift.internal.restclient.okhttp.OpenShiftAuthenticator$1.authenticate(OpenShiftAuthenticator.java:91)
at okhttp3.internal.http.RetryAndFollowUpInterceptor.followUpRequest(RetryAndFollowUpInterceptor.java:294)
at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:160)
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
at com.openshift.internal.restclient.okhttp.ResponseCodeInterceptor.intercept(ResponseCodeInterceptor.java:55)
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:121)
at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:200)
at okhttp3.RealCall.execute(RealCall.java:77)
at com.openshift.internal.restclient.okhttp.OpenShiftAuthenticator.tryAuth(OpenShiftAuthenticator.java:97)
at com.openshift.internal.restclient.okhttp.OpenShiftAuthenticator.authenticate(OpenShiftAuthenticator.java:59)
at okhttp3.internal.http.RetryAndFollowUpInterceptor.followUpRequest(RetryAndFollowUpInterceptor.java:294)
at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:160)
at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:147)
[...]

PR #358, opened 2018-12-07
using java.util.Base64 instead of javax.xml.bind.DatatypeConverter

adietish:
...to allow us to run on jdk9+
Signed-off-by: Andre Dietisheim <adietish@redhat.com>


adietish:
fixes #357

PR #361, opened 2019-02-06
Make sure http delete is using payload

### LINKED ISSUE #359: Payload Nullified when the request type is 'DELETE'
I notice that when deleting OpenShift Jobs from OpenShift web console it called:
  
| Name  | Value |
| ------------- | ------------- |
| URL  | https://$ENDPOINT/apis/batch/v1/namespaces/$NAMESPACE/jobs  |
| RequestMethod  | DELETE |
| Payload| {"kind":"DeleteOptions","apiVersion":"v1","propagationPolicy":"Foreground"} |

There is an inconsistency in the restclient API where it nullify any request with DELETE RequestMethod type. I understand some HTTP standards encourage to drop the payload on GET/DELETE although it is not explicitly stated here: https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html. 

https://github.com/openshift/openshift-restclient-java/blob/fee040f2872a406504a058fe67ab13ad4e061314/src/main/java/com/openshift/internal/restclient/DefaultClient.java#L316-L326

As much as I want to follow the common convention, I would prefer the GET/DELETE request method in this restclient not to drop the payload and let openshift endpoint to decide whether it will read the payload/ignore it. Otherwise, from openshift API side, the API needs to also follow the conventions by dropping the payload and open the parameter.

adietish:
fixes #359

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
+++ b/src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
@@ -101,7 +101,7 @@ public class ApiTypeMapper implements IApiTypeMapper, ResourcePropertyKeys {
                     .map(api -> formatEndpointFor(api, (split.length == 0 ? preferedVersion.get(api) : split[0]), kind))
                     .filter(e -> resourceEndpoints.contains(e)).findFirst();
         } else {
-            result = Optional.of(formatEndpointFor(API_GROUPS_API, version, ResourceKind.pluralize(kind, true, true)));
+            result = Optional.of(formatEndpointFor(API_GROUPS_API, version, kind));
         }
         if (result.isPresent()) {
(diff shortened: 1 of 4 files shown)


adietish:
fixes #359

adietish:
@jcantrill thx for the quick review!

adietish:
[merge]

adietish:
This change is causing problems when deleting resources on OpenShift. In JBoss Tools, using this version of the client library is resulting in the following error:
```
com.openshift.restclient.OpenShiftException: Exception trying to DELETE https://api.rh-us-east-1.openshift.com/api/v1/namespaces/testomat/pods/wfl14-7-wg76m response code: 500 no kind "Pod" is registered for version "v1" in scheme "k8s.io/apimachinery/pkg/apis/meta/internalversion/register.go:31"
```

Issue #371, opened 2019-02-13
[integration-tests] fix PodLogRetrievalAsyncIntegrationTest


adietish:
this is a subtask to #365

Issue #370, opened 2019-02-13
[integration-tests] fix OpenshiftBinaryRSyncRetrievalIntegrationTest


adietish:
this is a subtask #365

Issue #369, opened 2019-02-13
[integration-tests] fix OpenshiftBinaryPodLogRetrievalIntegrationTest


adietish:
this is a subtask #365

Issue #368, opened 2019-02-13
[integration-tests] fix AuthorizationKindsIntegrationTest


adietish:
this is a subtask to #365

PR #372, opened 2019-02-13
[365] fixes all ITs

adietish:
* #366
* #367
* #368 
* #369 
* #370 
* #371


src/test/java/com/openshift/internal/restclient/IntegrationTestHelper.java
```diff
@@ -75,14 +82,57 @@ public String getDefaultNamespace() {
...
-        return String.format("%s-%s", getDefaultNamespace(), new Random().nextInt(9999));
+        return appendRandom(getDefaultNamespace());
     }
 
     public IProject generateProject(IClient client) {
         IResource request = client.getResourceFactory().stub(ResourceKind.PROJECT_REQUEST, generateNamespace());
         return (IProject) client.create(request);
     }
 
+    public IProject getOrCreateProject(IClient client) {
```
jeffmaury: not sure to understand: if a project exist then you use this one bit 2 questions arise
- if the project gets deleted at test end then we will delete a user project
- if not then we will leave lots of resources on the cluster

adietish: @jcantrill you're right in pointing out this but I deliberately get away from following the basic pattern where a test would create resource and clean up afterwards. I try to reuse resources. My aim is to run the integration test against an instance that's as close as we can be to upstream OpenShift (origin). I thus target employee.openshift.com since it's regularly updated to the most recent version that's been released. I then also try to reuse resource as frequently as possible to address latency and IT execution time. I therefore try try to reuse resources, only creating them if they don't exist or don't offer characteristics that the test tries to assert. Back in OpenShift v2 with 100s of ITs this strategy turned out to be the only viable strategy, creating and reusing would have multiplied the immense time such integration tests would have taken to be executed.
Ideally of course we'd use a dedicated test user to run those tests.

jeffmaury: Yes but as ITs needs to be run as admin, then imagine what will happen to your cluster if one of the project you delete is one from OpenShift or Kube ? Why not create the project not before every test but when the test class is created ?

adietish: @jeffmaury what requires us to run the ITs as admin? 
I want to have these ITs running continuously on jenkins so that we get alarmed of server API breakage in the library. We could download & install a CDK onto jenkins to achieve this. Simpler would be to use employee.openshift.com. When running CDK locally I currently cannot run any > v3.5 so I run them against employee.openshift.com. On that cluster I cannot have >1 project, so creating a new project isn't possible, the test would have to kill the existing first which then is again moot.

jeffmaury: I think some of the IT (related to roles and policies) requires admin. But maybe these are the ones that are broken. Will try to run them against my local CDK without admin to see if this changes the results

adietish: @jeffmaury I think that you're right in the requirement of being admin. There's a command in the IT-properties file that adds the cluster role to the admin user:
```
#osadm policy add-cluster-role-to-user cluster-admin $ADMIN_USER --config=openshift.local.config/master/admin.kubeconfig
```
I'm not even sure if the osam command still exists, the current documentation refers to ``òc adm``` (see here: https://docs.openshift.com/container-platform/3.10/admin_guide/manage_rbac.html)

adietish: @jeffmaury so this leaves to with the following requirements:
* ability to run against latest OpenShift (in order to detect REST API breakage)
* ability to alter cluster policies (in order to be able to test roles, policies, etc.)

the 2nd requirement disallows us to use employee.openshift.com and to install a CDK or origin in jenkins.
Given this scenario I'd suggest that I dont use an existing project but make sure an integration-test project exists. Nevertheless I still believe that we should reuse resources and not clean up, to speed up test execution.

adietish:
@jeffmaury please re-review. The PR fixes all integration-tests. The main changes are the following:
*   [365] ignoring ITs for deprecated authorization endpoints
    * Role
    * PolicyBinding
    * Policy
* corrected rsync log assertion: removed é from filename since log reports
      name as octal utf8 by rsync
    * killing resources - not project - upon test teardown (project removal is
      async, prohibits new test from creating test resources)
    * wait for pod to be ready before testing against it
    * make sure new test resource has unique name (add random string to it)

adietish:
fixes 
* #366
* #367
* #368 
* #369 
* #370 
* #371

adietish:
@jeffmaury dare to review please? simply run `mvn clean verify -Pintegration-tests`

PR #375, opened 2019-03-13
allow configuration of ping interval and its time unit for OkHttpClie…

### LINKED ISSUE #374: Allow configuration of ping interval for OkHttpClient via ClientBuilder
The latest versions of OkHttpClient lib support the configuration of ping interval for websockets, but ClientBuilder API does not allow to specify ping interval. 
By default ping interval = 0, which means that no ping is performed by default. As the result websocket connection gets teared down by server and all registered IWatchers have to be reconnected and as side effect they load (on com.openshift.restclient.IOpenShiftWatchListener#connected event) all resources they are configured to watch for.
With non zero ping interval the connection can be keept alive.

denisanfimov:
Closes #374

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,7 +17,7 @@
     <!-- Artifact Information -->
     <groupId>com.openshift</groupId>
     <artifactId>openshift-restclient-java</artifactId>
-    <version>7.0.1-SNAPSHOT</version>
+    <version>7.1.0-SNAPSHOT</version>
     <packaging>jar</packaging>
     <name>OpenShift Java REST Client</name>
     <url>http://openshift.redhat.com</url>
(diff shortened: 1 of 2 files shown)


adietish:
@denisanfimov thx for pointing this out and even providing a PR. Awesome!
Given that we change the API we might want to bump the minor version 7.0.1->7.1.0. Agree?

denisanfimov:
Hello, André!
Thanks for quick response.
As for minor version change, yes I agree with you.

adietish:
@denisanfimov ok great, can you add this change please? Imho better than me bumping and then you rebasing.

denisanfimov:
done

adietish:
@denisanfimov great, thx!

adietish:
@denisanfimov deployed openshift-restclient-java 7.1.0-SNAPSHOT to the maven repository. Thanks!

PR #380, opened 2019-05-09
OSJC-284: Authorize URL is computed from baseURL use /.well-known/oau…

jeffmaury:
…th-authorization-server

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/DefaultClient.java
+++ b/src/main/java/com/openshift/internal/restclient/DefaultClient.java
@@ -498,14 +498,14 @@ public class DefaultClient implements IClient, IHttpConstants {
 
     private class AuthorizationCallback implements Callback {
 
-		private void setDefaults() {
-			DefaultClient.this.authorizationEndpoint.complete(DefaultClient.this.getDefaultAuthorizationEndpoint());
-        	DefaultClient.this.tokenEndpoint.complete(DefaultClient.this.getDefaultTokenEndpoint());
-		}
+        private void setDefaults() {
+            DefaultClient.this.authorizationEndpoint.complete(DefaultClient.this.getDefaultAuthorizationEndpoint());
+            DefaultClient.this.tokenEndpoint.complete(DefaultClient.this.getDefaultTokenEndpoint());
+        }
 
-		@Override
+        @Override
         public void onFailure(Call call, IOException e) {
-        	setDefaults();
+            setDefaults();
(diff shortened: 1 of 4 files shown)


src/main/java/com/openshift/internal/restclient/DefaultClient.java
```diff
@@ -504,6 +541,40 @@ public URL getBaseURL() {
...
+    protected URL getDefaultAuthorizationEndpoint() {
+        try {
+            return new URL(getBaseURL(), "oauth/authorize");
+        } catch (MalformedURLException e) {
+            throw new OpenShiftException(e, e.getLocalizedMessage());
+        }
+    }
+
+    @Override
+    public URL getTokenEndpoint() {
```
jcantrill: This PR looks good to me.  Only comment I would have is if we desire to use URL.  I recall @adietish preferring string because the JVM doesn't try to resolve the endpoint or something to that affect?

jeffmaury: I don't think JVM is resolving the endpoint when creating the URL object. It just parses the string to check for correct syntax. The only drawback is that checked exception is thrown by the constructor but I kept URL to be inline with `getBaseURL`

jeffmaury:
@adietish dare to review and merge ?

src/main/java/com/openshift/internal/restclient/DefaultClient.java
adietish: @jeffmaury @jcantrill I generally try to avoid URL bcs of the following: the problem with URL is that it resolves hostnames to IP addresses in #equals and #hashCode.
```
    protected synchronized InetAddress getHostAddress(URL u) {
        ...
        try {
                u.hostAddress = InetAddress.getByName(host);
        } catch (UnknownHostException ex) {
                return null;
        } catch (SecurityException se) {
        ...

```
This is generally a bad idea given that these methods should be consistent which they can't if they rely on an unreliable resource like the internet.
As a rule of thumb it's therefore preferable to use URI.

A quick google showed the following stackoverflow which summarizes the problem: https://stackoverflow.com/a/2349535/231357

Given that we're not using #equals here ourself (by ex. putting it into a hashmap) and we're already offering 
```
    @Override
    public URL getBaseURL() {
```
 I think that it's ok and consistent to use URL here.

adietish:
@jeffmaury deployed 7.1.1-SNAPSHOT to repository.jboss.org

PR #389, opened 2019-06-21
bumping to 8.0.0.Final

adietish:


jeffmaury:
I would suggest that we postpone a little bit the merge so that we can monitor the latest JBoss Tools integration tests and check if we can observe regression.
We can target late Monday or early Tuesday if everything is ok.
WDYT @adietish ?

jeffmaury:
Checked the latest JBoss Tools integration tests and did not see any regressions. So we're good to merge.

adietish:
Artifact is deployed to https://repository.jboss.org/nexus/service/local/repositories/releases/content/com/openshift/openshift-restclient-java/8.0.0.Final/openshift-restclient-java-8.0.0.Final.jar
Master is tagged.

PR #393, opened 2019-07-09
NPE

m-pavel:
DeploymentRequest is registered as v1.DeploymentRequest in ITypeFactory, so it can not be found by name DeploymentRequest


adietish:
@m-pavel thx for patch! Can you provide steps so that I can reproduce the bug, please?

PR #401, opened 2019-10-03
impl'd capability to retrieve console url (#400)

### LINKED ISSUE #400: Add capability to retrieve url for the web-console
We currently have this logic in JBoss Tools. With the difference in OpenShift 4 (OpenShift 3 has the console on the same host as the REST endpoint. OpenShift 4 does not have this restriction, it has a config map pointing to it) we should now move this logic to the openshift-restclient-java.

adietish:
fixes #400

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/capability/server/Console.java
+++ b/src/main/java/com/openshift/internal/restclient/capability/server/Console.java
@@ -1,0 +1,189 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Red Hat, Inc. Distributed under license by Red Hat, Inc.
+ * All rights reserved. This program is made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution, and is
+ * available at http://www.eclipse.org/legal/epl-v10.html
+ * 
+ * Contributors: Red Hat, Inc.
+ ******************************************************************************/
+
+package com.openshift.internal.restclient.capability.server;
+
+import java.util.Arrays;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+import com.openshift.restclient.IClient;
+import com.openshift.restclient.ResourceKind;
+import com.openshift.restclient.capability.server.IConsole;
+import com.openshift.restclient.model.IConfigMap;
+import com.openshift.restclient.model.IProject;
+import com.openshift.restclient.model.IResource;
+
+public class Console implements IConsole {
+
+    public static final String CONFIGMAP_DATA_CONSOLE_URL = "consoleURL";
+    public static final String NAMESPACE_OPENSHIFT_CONFIG_MANAGED = "openshift-config-managed";
+    public static final String CONFIGMAP_CONSOLE_PUBLIC = "console-public";
+
+    private IClient client;
+
+    public Console(IClient client) {
+        this.client = client;
+    }
+
+    @Override
+    public boolean isSupported() {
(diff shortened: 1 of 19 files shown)


src/main/java/com/openshift/internal/restclient/capability/server/Console.java
```diff
@@ -0,0 +1,186 @@
...
+        String projectName = resource == null ? null : resource.getNamespaceName();
+        if (projectName != null) {
+            builder.append("/project/").append(projectName);
+        }
+        if (resource != null 
+                && !(resource instanceof IProject)) {
+            String consoleResourceUrl = getOpenShiftResourceURL(resource, 
+                    Arrays.stream(OpenShiftConsoleResourceUrls.values())
+                            .filter(resourceUrl -> resourceUrl.getVersion() ==  3));
+            if (consoleResourceUrl != null) {
```
jeffmaury: So if this is called for a resource type not supported by OpenShiftConsoleResourceUrls then null is returned but the builder has some kind of incomplete URL, is this on purpose ?

src/test/java/com/openshift/internal/restclient/capability/resources/BuildCapabilitiesIntegrationTest.java
```diff
@@ -57,14 +57,14 @@ public void setUp() throws Exception {
         // a buildconfig
         IBuildConfigBuilder builder = client.adapt(IBuildConfigBuilder.class);
         assertNotNull("Exp. the client to be able to use a buildconfigbuilder", builder);
-        LOG.debug("Creating BuildConfig {}", bc);
+        LOG.debug("Creating BUILDCONFIG {}", bc);
```
jeffmaury: What's the rationale for that ?

src/main/java/com/openshift/internal/restclient/capability/server/Console.java
adietish: if a resource type is not supported, then we should show the project overview. I added test (data points) that assert this.

src/test/java/com/openshift/internal/restclient/capability/resources/BuildCapabilitiesIntegrationTest.java
adietish: None, that's an unintentional change. Thanks for spotting. I'm correcting it.

adietish:
@jeffmaury 
> I would really like to see an integration test for this new capability
I added integration tests for the openshift major version and the console url. I added a simple test where I get the console url and http get it. Anything that goes beyond this is hard to test though:
* console root OS3: reports 403 for invalid URLs -> requesting and asserting 200 OK, works
* resources OS3: reports 200 for inexistant resource with an error in html as long as there's js enabled. If not it's complaining about missing js enablement -> dont see how I would test this 
* console root OS4: reports 200 for invalid URLs -> dont see how I would test this
* resources OS3: reports 200 for inexistant resource with an error in html as long as there's js enabled. If not it's complaining about missing js enablement -> dont see how I would test this

adietish:
[merge]

adietish:
openshift-restclient-java 8.1.0-SNAPSTHO deployed to repository.jboss.org at https://repository.jboss.org/nexus/service/local/repositories/snapshots/content/com/openshift/openshift-restclient-java/8.1.0-SNAPSHOT/openshift-restclient-java-8.1.0-20191004.212928-1.jar

PR #411, opened 2019-10-25
authorize by direct request to auth-endpoint (was: using invalid bearer & authenticator indirection)

adietish:
The change here is all requests hold authentication means. If those are not present an unauthenticated request to the authentication endpoint is performed in order to retrieve the bearer token. Before this fix, the request to the authentication endpoint was holding an invalid bearer token (which was ignored in 3.x but causes a Resource Forbidden Exception in CRC/OS 4.2). This causes that authentication is performed.

The bug that's reporting this issue is filed in JBoss Tools: https://issues.jboss.org/browse/JBIDE-26899

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/okhttp/AuthenticatorInterceptor.java
+++ b/src/main/java/com/openshift/internal/restclient/okhttp/AuthenticatorInterceptor.java
@@ -13,8 +13,6 @@ package com.openshift.internal.restclient.okhttp;
 
 import java.io.IOException;
 import java.net.URL;
-import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Map;
 
 import org.apache.commons.lang.StringUtils;
@@ -29,13 +27,11 @@ import com.openshift.restclient.authorization.IAuthorizationDetails;
 import com.openshift.restclient.authorization.UnauthorizedException;
 import com.openshift.restclient.http.IHttpConstants;
 
-import okhttp3.Authenticator;
 import okhttp3.Interceptor;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Request.Builder;
 import okhttp3.Response;
-import okhttp3.Route;
 
 /**
  * Adds authorization means to every request. Authorizes and retrieves the token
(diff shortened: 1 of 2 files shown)


src/main/java/com/openshift/internal/restclient/okhttp/OpenShiftRequestBuilder.java
```diff
@@ -57,9 +57,9 @@ public OpenShiftRequestBuilder authorization(IAuthorizationContext authorization
     }
 
     public OpenShiftRequestBuilder authorization(String token) {
-        // Authenticator only gets triggered with a 401.
-        // We thus always use the token even if it's null so that we get a 401 instead of 403 in OS 4.2
-        builder.header(IHttpConstants.PROPERTY_AUTHORIZATION, IHttpConstants.AUTHORIZATION_BEARER + " " + token);
+        if (!StringUtils.isBlank(token)) {
```
adietish: explanation: dont add a AUTHORIZATION_BEARER header if we have no token (before: was adding an invalid bearer with the value "null")

src/main/java/com/openshift/internal/restclient/okhttp/AuthenticatorInterceptor.java
```diff
@@ -84,21 +79,21 @@ private boolean isUrlWithoutAuthorization(Request request, String url) {
...
-                .addHeader(CSRF_TOKEN, "1")
-                .url(new URL(client.getAuthorizationEndpoint().toExternalForm() 
-                        + "?response_type=token&client_id=openshift-challenging-client").toString())
-                .build();
+        Request authRequest = appendAuthorization(
+                client.getAuthorizationContext(),
+                new Request.Builder()
+                    .addHeader(CSRF_TOKEN, "1")
+                    .url(new URL(client.getAuthorizationEndpoint().toExternalForm() 
+                            + "?response_type=token&client_id=openshift-challenging-client").toString()));
```
adietish: explanation: request the bearer token without authenticator which is an unnecessary indirection)

adietish:
@jeffmaury please review & add /lgtm label

adietish:
@jeffmaury rebased, please review & add /lgtm label

adietish:
@jeffmaury plase review and add /lgtm label

PR #427, opened 2020-01-28
removed unnecessary param request when testing if url needs auth

adietish:


adietish:
@jeffmaury dare to please verify this simple cleanup?

PR #431, opened 2020-01-30
correctly handle challenge headers with 'realm=' (#430)

### LINKED ISSUE #430: BasicChallengeHandler#canHandle false negative when value contains "realm"
Current impl for BasicChallengeHandler#canHandle is returning false if the response header is using ```realm=```:
```
WWW-Authenticate: Basic realm="smurfs"
```

adietish:
fixes #430


adietish:
@jeffmaury can you please review this and /lgtm if you agree with this change?

PR #434, opened 2020-02-07
Bump sl4j api version

adietish:
depends on #432 which has to be merged first


adietish:
@jeffmaury dare to please give me an /lgtm to this simple change in version of the sl4j-api (forgot that one when bumping log4j)?

adietish:
@jeffmaury it's required so that it's on par with the other artifacts of the log4j/sl4j combo. 2.0.0.-Alpha1  is a new release since ages (1.7 is 14y old) and seems to be what quite a few projects bet on nowadays.

PR #436, opened 2020-02-13
only log to slf4j (use log4j2 only in tests, integration-tests)

adietish:


adietish:
@jeffmaury removed slf4j which turned out to carry its own set of problems. This is required for https://github.com/jbosstools/jbosstools-openshift/pull/1908. Please review.

adietish:
@jeffmaury please review and /lgtm

adietish:
> This PR is broken: you use SLF4J only in pom.xml but import for Log4J are still there in code.
> 
> Another question: because you don't have any SLF4J provider, what about the tests ? Do we see the messages ?

adietish:
> This PR is broken: you use SLF4J only in pom.xml but import for Log4J are still there in code.

ouch, my bad, incomplete commit here. Fixing. 

> Another question: because you don't have any SLF4J provider, what about the tests ? Do we see the messages ?
The beauty of only having slf4j is that it's up to the consumer to decide what to use. The library does not impose a choice, only provide the facade. Kubernetes-client does the same. 
For the integration-tests, yes, that's an issue. We should find a way to only provide log4j for those tests.

adietish:
deployed artifact to https://repository.jboss.org/nexus/service/local/repositories/snapshots/content/com/openshift/openshift-restclient-java/8.1.1-SNAPSHOT/openshift-restclient-java-8.1.1-20200218.131349-19.jar

Issue #456, opened 2020-07-20
Update resource cannot support Chinese characters


a85302543:
When I use Openshift Java Client to create or update a resource , I found a problem  :when  configuring  env with Chinese character like that: 
```java
"env": [
  {
    "name": "key-2",
    "value": "中文字符"
  }, 
  {
   "name": "key-1",
    "value": "value-1"
  }
]

```
and the return values is wrong and like:
```java
"env": [
  {
    "name": "key-2",
    "value": �����������������
  }, 
  {
   "name": "key-1",
    "value": "value-1"
  }
]
```
The problem seems like that when posting the request, the Chinese character can not be recognized . A temporary solution is changing the code.
  from 
```java
 return RequestBody.create(json, MediaType.parse("application/json"));
```
to 
```java
 return RequestBody.create(json, MediaType.parse("application/json;charset=utf-8"));
```

is there other way to avoid this or this problem can be sovled :) thanks
```Java
Version: 9.0.0.Final
Class:DefalutClient.java
line: 339
```

adietish:
Hi

I think that there's no harm in what you suggest and should be included if it fixes your issue. Dare to suggest a PR, please?
@jeffmaury Thoughts?

adietish:
@a85302543 let me please point out that this client is in maintenance mode, we'll only fix existing/new bugs as far as they are essential to the proper functioning of this library. If you're looking for a more feature complete & active library, we suggest that you use https://github.com/fabric8io/kubernetes-client/ instead.

a85302543:
@adietish Hi 
thanks for your reply！ I'll give it a try

adietish:
Fix at #459 merged. Closing.

PR #459, opened 2020-07-23
Fix (#456)

a85302543:
update resource support Chinese character (#456)


adietish:
@a85302543 thanks for the fix!

Issue #454, opened 2020-06-25
Raw Execution Request for DELETE method with null payload is causing a problem


alfonzjanfrithz:
To reproduce this issue:

```java
IClient client = new ClientBuilder("...").build();
client.execute(HttpMethod.DELETE.name(), "Job", "batch/v1", "test-namespace", "pi", null, null, null);
``` 

This leads to a the following problem:
```log
java.lang.IllegalArgumentException: Parameter specified as non-null is null: method okio.Okio.source, parameter $receiver
	at okio.Okio.source(Okio.kt)
	at com.openshift.internal.restclient.DefaultClient$1.writeTo(DefaultClient.java:351)
	at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.kt:59)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:112)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:87)
	at com.openshift.restclient.ClientBuilder$UserAgentInterceptor.intercept(ClientBuilder.java:390)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:112)
	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.kt:37)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:112)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:87)
	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:82)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:112)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:87)
	at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.kt:84)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:112)
	at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.kt:71)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:112)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:87)
[...]

PR #460, opened 2020-07-27
Ensure null payload (on DELETE method) is transformed into empty string

alfonzjanfrithz:
This is the code change to fix issues [454](https://github.com/openshift/openshift-restclient-java/issues/454)

**What's changed**
- There will be a check if the payload used by `client.execute()` containing null values for `InputStream` parameter.
- Refactoring typo in the test
- Access modifier for `DefaultClient.getPayload()` changed to default to be accessible by the test


adietish:
@alfonzjanfrithz thanks a lot, looks good!

PR #464, opened 2020-08-04
expose IAuthorizationContext#expires (#463)

### LINKED ISSUE #463: When authenticating, "expiresIn" is not set
in https://github.com/openshift/openshift-restclient-java/issues/455#issuecomment-663590391 we discovered that ```IAuthenticationContext.getExpiresIn()``` always returns null and obviously isn't filled when authenticating.
The response carries the value though:

```
Location: https://oauth-openshift.apps.dev.xxxxxxxxxxx/oauth/token/implicit#access_token=token-here-but-omitted&expires_in=7200&scope=user%3Afull&token_type=Bearer
```

adietish:
fixes #463

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/authorization/AuthorizationContext.java
+++ b/src/main/java/com/openshift/internal/restclient/authorization/AuthorizationContext.java
@@ -9,9 +9,12 @@
 
 package com.openshift.internal.restclient.authorization;
 
+import java.time.Instant;
+
 import org.apache.commons.lang.StringUtils;
 
 import com.openshift.restclient.IClient;
+import com.openshift.restclient.OpenShiftException;
 import com.openshift.restclient.ResourceKind;
 import com.openshift.restclient.authorization.IAuthorizationContext;
 import com.openshift.restclient.authorization.IAuthorizationDetails;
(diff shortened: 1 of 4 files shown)


adietish:
@mhagnumdw could you please test this? Would love to get your feedback. I added ```IAuthorizationContext.getExpiresIn()``` & ```IAuthorizationContext.getExpires()```. The prior returns the value that was included in the authorization response while the latter would indicate the moment at which the token expires (adds expiresIn to the creation timestamp).

mhagnumdw:
@adietish, I ran the test and everything looks fine!

I'll just add a note at one point in the code in case you think it's worth handling a `NumberFormatException`.

Tks!!

adietish:
> @adietish, I ran the test and everything looks fine!
> 
> I'll just add a note at one point in the code in case you think it's worth handling a `NumberFormatException`.
> 
> Tks!!

Ideally we should throw an OpenShift exception in this case specifying that we could not determine the expires timestamp. Agree?

mhagnumdw:
> Ideally we should throw an OpenShift exception in this case specifying that we could not determine the expires timestamp bcs the expiresIn wasn't in the correct format. Anything else doesn't seem to make any sense. Agree?

Sounds good.

adietish:
Hi @mhagnumdw

I am now throwing an OpenShiftException when "expiresIn" has an illegal value and cannot be parsed. 
Can I get your +1 or dissenting opinion, please?

adietish:
[merge]

adietish:
merged.

PR #465, opened 2020-08-10
corrected logging for inaccessible legacy apis to info (#462)

### LINKED ISSUE #462: log: Can't access legacy endpoint oapi
Hi !

I get this log in my application on the first OpenShift request: `2020-08-03 18:20:25.428 ERROR com.openshift.internal.restclient.ApiTypeMapper:251 (qtp1327536153-58) - Can't access legacy endpoint oapi`

It appears here https://github.com/openshift/openshift-restclient-java/blob/6e4454dc50640a656ebed46bd85579b59ab4ddce/src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java#L251

Should it really be the `ERROR` level?

adietish:
fixes #462

### DIFF
--- a/src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
+++ b/src/main/java/com/openshift/internal/restclient/ApiTypeMapper.java
@@ -242,13 +242,13 @@ public class ApiTypeMapper implements IApiTypeMapper, ResourcePropertyKeys {
 
     private Collection<ApiGroup> getLegacyGroups() {
         Collection<ApiGroup> groups = new ArrayList<>();
-        for (String e : Arrays.asList(KUBE_API, OS_API)) {
+        for (String endpoint : Arrays.asList(KUBE_API, OS_API)) {
             try {
-                String json = readEndpoint(e);
+                String json = readEndpoint(endpoint);
                 ModelNode n = ModelNode.fromJSONString(json);
-                groups.add(new LegacyApiGroup(e, n));
+                groups.add(new LegacyApiGroup(endpoint, n));
             } catch (Exception ex) {
-                LOGGER.error("Can't access legacy endpoint {}", e);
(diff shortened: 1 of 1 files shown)


adietish:
Hi @mhagnumdw

Can I please have your +1?

mhagnumdw:
Hi @adietish, the only slightly different thing I saw is that you use the exception `toString()`. When we usually use `getMessage()` or even delegate the exception instance to the logger, something like `logger.info(String msg, Throwable t)` (but it's strange INFO prints an exception).

If this was not by mistake, that's fine by me! Tks!!

adietish:
> Hi @adietish, the only slightly different thing I saw is that you use the exception `toString()`. When we usually use `getMessage()` or even delegate the exception instance to the logger, something like `logger.info(String msg, Throwable t)` (but it's strange INFO prints an exception).

Oh, nice catch, didn't notice. This is clearly by mistake, caused by misinterpretation of the constructor parameters of OpenShiftException. Fully agree on the #getMessage to report the bottom line of the exception. On the other hand, the original exception should be kept and passed along when wrapping. I'm correcting it.

mhagnumdw:
Ohh.. Sorry! My mistake! I got confused when reading the code.

mhagnumdw:
I tested it locally... Before the exception was not put in the log. Now, the exception is placed in the log and as INFO. An exception printed as INFO is strange. Well... feel free to approve the MR and thanks again!

I hope I didn't read the code wrong again 😅

adietish:
> I tested it locally... Before the exception was not put in the log. Now, the exception is placed in the log and as INFO. An exception printed as INFO is strange. 

Yes I agree, but then it's quite useless to get a message that the oapi could not be reached without knowing the reason for it.

adietish:
merged.

PR #478, opened 2022-12-13
Delete .travis.yml

coreydaley:


adietish:
@coreydaley thanks for noticing and taking care :)

PR #480, opened 2023-04-05
Refactor: Refactoring code

Saumya-09:
1. GetEnvSuper.java : Both the interfaces ISourceBuildStrategy and IJenkinsPipelineStrategy had the same method getEnvVars(), so I created a new interface and implemented that directly where getEnvVars() was used.
2. SourceBuilderStrategy.java : Renamed the variable to avoide confusion with reserved keyword "var".
3. SourceBuilderStrategy.java and JenkinsPipelineStrategy : Both the method had exactly same method implementaion, so created a new class with the method that can be reused.
4. ApiTypeMapper : Decomposed the condition to make it more readable and maintainable.
5. ServiceAccount.java : Two methods had exactly same implementation so extracted the method to avoide duplication of code.


adietish:
@Saumya-09 Thanks for your contribution. But this client isn't maintained any longer. Please consider moving to the newer and actively developed https://github.com/fabric8io/kubernetes-client which includes an openshift-client, too.

